PR #4, opened 2020-04-28
New: Add project dockerized platform

pedroamador:


Jenkinsfile
```diff
@@ -60,14 +60,18 @@ pipeline {
...
-                docker {
-                    image 'maven:3.6.3-jdk-11'
-                    label 'docker'
-                }
-            }
+            agent { label 'docker'}
             steps {
-                sh 'bin/devcontrol.sh backend integration-tests'
+                script {
+                    docker.image('docker:dind').withRun('--privileged -v "$WORKSPACE":"$WORKSPACE" --workdir "$WORKSPACE"') { c ->
```
ibaiul, line 66: Wouldn't it be better if we create a docker image with all the dependencies we need and upload it to docker hub instead or creating one from scratch every time we launch the pipeline?

pedroamador: Yes, you're right. It's better to have a Dockerfile and build && push a docker image (maybe in a previous step).

I will change that shortly

PR #12, opened 2020-05-23
How to run the API from terminal with maven

RubenZagon:


ibaiul (changes requested):
1- In the section where it explains how to run the tests, maybe it is worth mentioning that the docker daemon must already be running in your system in order to run the integrations tests that require it.

2- Notice that the platform can also be run directly from the IDE by just right clicking in the Application.java file which might be more convenient in some cases.

3- I would mention as well that before running the the backend locally the docker-compose file under backend/docker/local directory must be up to provide the database/localstack dependencies

RubenZagon:
> I would mention as well that before running the the backend locally the docker-compose file under backend/docker/local directory must be up to provide the database/localstack dependencies

Very nice observation @ibaiul thanks for the feedback, **i did the changes you comment** 🚀

carlosble (approved):

ibaiul (approved):

PR #16, opened 2020-06-04
Feature/enable cors

RubenZagon:
Use a global configuration to CORS.

we enable the origin localhost:3000 to comunication with the front in develop, **i don't know if we need another origin in producction**

### DIFF
--- a/backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
+++ b/backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
@@ -2,6 +2,7 @@ package com.huellapositiva.infrastructure;
 
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.http.HttpMethod;
 import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
@@ -20,6 +21,9 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 @EnableGlobalMethodSecurity(jsr250Enabled = true)
 class SecurityConfig extends WebSecurityConfigurerAdapter {
 
+    @Value("${cors.allow.origin}")
+    private String origin;
+
     @Bean
     PasswordEncoder passwordEncoder() {
         return new BCryptPasswordEncoder();
@@ -31,8 +35,7 @@ class SecurityConfig extends WebSecurityConfigurerAdapter {
             @Override
             public void addCorsMappings(CorsRegistry registry) {
                 registry.addMapping("/api/v1/**")
-                        .allowedOrigins("http://localhost:3000")
-                        .allowedMethods("GET", "POST", "PUT", "DELETE")
+                        .allowedOrigins(origin)
                         .maxAge(3600);
             }
         };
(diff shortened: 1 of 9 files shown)


backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
```diff
@@ -23,12 +25,25 @@ PasswordEncoder passwordEncoder() {
         return new BCryptPasswordEncoder();
     }
 
+    @Bean
+    public WebMvcConfigurer corsConfigure(){
+        return new WebMvcConfigurer() {
+            @Override
+            public void addCorsMappings(CorsRegistry registry) {
+                registry.addMapping("/api/v1/**")
+                        .allowedOrigins("http://localhost:3000")
```
ibaiul: Externalize this value to a property that can be set in the different environments or it will not work in DEV and PROD

backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
```diff
@@ -23,12 +25,25 @@ PasswordEncoder passwordEncoder() {
...
+                        .allowedMethods("GET", "POST", "PUT", "DELETE")
```
ibaiul: Is it worth allowing HEAD and OPTIONS methods? Check if your frontend js framework requires it, I think Angular for instance needed it

backend/src/test/java/com/huellapositiva/integration/AwsEmailServiceShould.java
```diff
@@ -18,7 +18,7 @@
 @ExtendWith(AwsEnvVariablesExtension.class)
 @ExtendWith(LocalstackDockerExtension.class)
 @LocalstackDockerProperties(services = { "ses" })
-@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.NONE, properties = {"huellapositiva.feature.email.enabled=true"})
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT, properties = {"huellapositiva.feature.email.enabled=true"})
```
ibaiul, line 21: Why is that necessary?

ibaiul (changes requested):

backend/src/test/java/com/huellapositiva/integration/AwsEmailServiceShould.java
josnih21: I was having some issues related with AWS Service Email Test so I figure out why this test has not a defined port, when I change this property with the rest of Cors config enabled this test works again. 

I'm not sure at all why but this change makes test passed with cors config enabled

josnih21: I found this when executing webEnvironment to .NONE throws an exception saying that Its necesary to define this **Consider defining a bean named mvcHandlerMappingIntrospector in your configuration.**
 So that was the reason of changing to DEFINED_PORT

backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
RubenZagon: Okay, done.

backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
RubenZagon: It's not necessary for react, u are right. Remove the allowedMethods

backend/src/test/java/com/huellapositiva/integration/AwsEmailServiceShould.java
ibaiul: Miraremos eso en el futuro

ibaiul (approved):

PR #17, opened 2020-06-07
Tell don't ask

seymourpoler:
Hello Ayuda Digital,

I''ve applied tell don't ask principle in Email domain object, because, I think that Email has not to know the internal dependencies of EmailConfirmation.

### DIFF
--- a/backend/src/main/java/com/huellapositiva/domain/Email.java
+++ b/backend/src/main/java/com/huellapositiva/domain/Email.java
@@ -15,7 +15,7 @@ public class Email {
     private String body;
 
     public static Email createFrom(EmailConfirmation emailConfirmation, EmailTemplate emailTemplate, String from) {
-        if (emailConfirmation.getEmailAddress().isEmpty()) {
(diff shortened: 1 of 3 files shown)


ibaiul (changes requested):
The @NotEmpty validation of an email address is always performed at the controller level and should not be revalidated again. Only additional business related validations add some value.

In fact, the Email.from method should not even take an EmailConfirmation object as a parameter, but an EmailAddress object instead.

RubenZagon:
I will close this PR because we are using for now a Trello to work in the features of the pull request, if you wanna join us @seymourpoler you can contact with us, for know how we work in this moment of development.

Thanks again for you time :)

carlosble:
Hi! our current development branch is still having this code smell, the validation has not been removed. So we should not just close the PR but to fix the issue in one way or another. Also it would be helpful if you could post the link to the Trello over here so that @seymourpoler may ask for permission to join the team. Thank you!  @ibaiul @RubenZagon

PR #19, opened 2020-06-07
configure source and target maven compiler for the project

seymourpoler:

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -295,6 +295,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>${maven.compiler.version}</version>
         <configuration>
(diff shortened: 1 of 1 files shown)


ibaiul:
That is already the default implicit behaviour, no need to add it.
See: https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html

carlosble:
Hi! @seymourpoler if your proposal fixes an issue on your machine, please let us know as we all may have slightly different configurations and we want the project to be easy to set up for everybody. So if adding this lines is not but and moreover, made the project compile for you, then I will add them to the project @ibaiul . Thanks for collaborating with us :-)

PR #23, opened 2020-06-09
Feature/resend email confirmation

alu0100897562:
We need a review of the code to be sure everything is ok


ibaiul:
Must not include commits from any feature that is still under development

PR #31, opened 2020-06-15
Feature/login in register

ibaiul:


Nachocss (changes requested):
To fix: Don't send error trace to frontend when an exception occurs

Nachocss:
In JwtControllerShould, we should assert that the new tokens are not null, since asserting that a null token is not equal to the old one would pass the test correctly.

Nachocss (approved):

PR #36, opened 2020-06-24
Move catch responsibility to VolunteerService

ManuPerezDev:


RubenZagon (approved):

backend/src/main/java/com/huellapositiva/domain/service/VolunteerService.java
```diff
@@ -21,8 +23,12 @@
...
     public Integer registerVolunteer(PlainPassword plainPassword, EmailConfirmation emailConfirmation) {
-        PasswordHash hash = new PasswordHash(passwordEncoder.encode(plainPassword.toString()));
-        ExpressRegistrationVolunteer expressVolunteer = new ExpressRegistrationVolunteer(hash, emailConfirmation);
-        return volunteerRepository.save(expressVolunteer);
+        try {
+            PasswordHash hash = new PasswordHash(passwordEncoder.encode(plainPassword.toString()));
+            ExpressRegistrationVolunteer expressVolunteer = new ExpressRegistrationVolunteer(hash, emailConfirmation);
+            return volunteerRepository.save(expressVolunteer);
+        } catch (DataIntegrityViolationException ex) {
+            throw new UserAlreadyExists();
```
ibaiul, line 31: This exception should take a message and the original exception as constructor parameters to avoid losing the context in case the exception occurs due to a bug. DataIntegrityViolation could mean that other unique constraints are being violated apart of the email-address/username one

PR #37, opened 2020-06-26
Security/CSRF token is now required except on Login/Register/Health and GET requests

Nachocss:


backend/src/main/java/com/huellapositiva/application/controller/VolunteerApiController.java
```diff
@@ -40,7 +40,7 @@ public JwtResponseDto registerVolunteer(@Validated @RequestBody CredentialsVolun
             return jwtService.create(username, roles);
         } catch (PasswordNotAllowed pna) {
             throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Password doesn't meet minimum length");
-        } catch (UserAlreadyExists ex) {
+        } catch (FailedToPersistUser ex) {
             throw new ResponseStatusException(HttpStatus.CONFLICT, "User already exists");
```
ibaiul: El mensaje de la excepcion no es fiel a la realidad. podria haber fallado por otro motivo que no sea ese

backend/src/test/java/com/huellapositiva/api/EmailAddressConfirmationControllerShould.java
```diff
@@ -60,10 +62,15 @@ void confirm_email_with_non_existing_hash_should_return_404() throws Exception {
...
+
 
         // WHEN + THEN
         mvc.perform(post(baseUri + "/resend-email-confirmation")
                 .with(user(withMockUser(DEFAULT_EMAIL, VOLUNTEER_NOT_CONFIRMED)))
+                .header("X-XSRF-TOKEN", xsrfToken)
+                .cookie(loginResponse.getCookie("XSRF-TOKEN"))
                 .contentType(APPLICATION_JSON))
                 .andExpect(status().isNoContent());
     }
```
ibaiul, line 75: Faltan tests que demuestren que si no enviamos el token CSRF se nos deniega el acceso

backend/src/test/java/com/huellapositiva/util/TestUtils.java
```diff
@@ -49,24 +49,17 @@ public static void anonymousPostRequest(MockMvc mvc, String path, String body, R
...
 
-    public static JwtResponseDto refreshRequest(MockMvc mvc, String refreshToken) throws Exception {
-        String jsonResponse =  mvc.perform(post("/api/v1/refresh")
-                .content(refreshToken)
-                .contentType(APPLICATION_JSON)
-                .accept(APPLICATION_JSON))
-                .andExpect(status().isOk())
-                .andReturn().getResponse().getContentAsString();
-        return objectMapper.readValue(jsonResponse, JwtResponseDto.class);
+    public static String getTokenValue(String xsrfToken) {
```
ibaiul: el metodo se deberia llamar getCsrfTokenFromCookieHeader(String cookieHeader)

ibaiul (changes requested):

backend/src/main/java/com/huellapositiva/domain/service/VolunteerService.java
```diff
@@ -28,7 +30,8 @@ public Integer registerVolunteer(PlainPassword plainPassword, EmailConfirmation
             ExpressRegistrationVolunteer expressVolunteer = new ExpressRegistrationVolunteer(hash, emailConfirmation);
             return volunteerRepository.save(expressVolunteer);
         } catch (DataIntegrityViolationException ex) {
-            throw new UserAlreadyExists();
+            log.error("Unable to persist volunteer due to a conflict.", ex);
```
ibaiul, line 33: If we do not log the error we wont be able to identify issues at the time of creating a user.

backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
```diff
@@ -54,7 +55,11 @@ public void addCorsMappings(CorsRegistry registry) {
 
     @Override
     protected void configure(HttpSecurity http) throws Exception {
-        http.cors().and().csrf().disable().authorizeRequests()
+        http.cors().and()
+                .csrf()
+                    .ignoringAntMatchers("/api/v1/volunteers/login", "/api/v1/volunteers/register")
```
ibaiul, line 60: GET requests are allowlisted by default, no need any logic for that.

backend/src/test/java/com/huellapositiva/api/EmailAddressConfirmationControllerShould.java
```diff
@@ -60,10 +64,12 @@ void confirm_email_with_non_existing_hash_should_return_404() throws Exception {
     void resend_email_should_return_204() throws Exception {
         //GIVEN
         testData.createCredential(DEFAULT_EMAIL, UUID.randomUUID(), DEFAULT_PASSWORD, VOLUNTEER_NOT_CONFIRMED);
+        MockHttpServletResponse loginResponse = loginRequest(mvc, new CredentialsVolunteerRequestDto(DEFAULT_EMAIL, DEFAULT_PASSWORD));
 
         // WHEN + THEN
         mvc.perform(post(baseUri + "/resend-email-confirmation")
                 .with(user(withMockUser(DEFAULT_EMAIL, VOLUNTEER_NOT_CONFIRMED)))
+                .with(csrf())
```
ibaiul, line 72: In tests where we do not want to test CSRF explicitly it is better to simulate it through Spring test utils to have a cleaner test that focuses only on the task we want to achieve. The same applies for logged in users.

backend/src/test/java/com/huellapositiva/api/SecurityConfigShould.java
```diff
@@ -0,0 +1,126 @@
...
+import static com.huellapositiva.util.TestUtils.loginRequest;
+import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.springframework.http.MediaType.APPLICATION_JSON;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@AutoConfigureMockMvc(printOnlyOnFailure = false)
```
ibaiul, line 29: THIS DOES THE TRICK!! Don't ask me why, it makes no sense. If it fails in the future we could consider it a valid case to force dirty context refresh since we do want to use csrf() in the rest of the tests and that apparently creates session attributes that cannot be cleaned easily.

backend/src/test/java/com/huellapositiva/api/SecurityConfigShould.java
```diff
@@ -0,0 +1,126 @@
...
+                .andExpect(status().isForbidden());
+    }
+
+    @Test
+    void deny_access_when_csrf_token_is_invalid() throws Exception {
+        // WHEN + THEN
+        mvc.perform(post(REFRESH_URL)
+                .contentType(APPLICATION_JSON)
+                .header("X-XSRF-TOKEN", "invalid token")
+                .cookie(new Cookie("XSRF-TOKEN", "invalidCookie"))
```
ibaiul: Token and cookie consist in the same value

backend/src/test/java/com/huellapositiva/api/SecurityConfigShould.java
```diff
@@ -0,0 +1,126 @@
...
+                .andExpect(status().isForbidden());
+    }
+
+    @Test
+    void allow_access_when_csrf_token_is_valid() throws Exception {
+        // GIVEN
+        testData.createVolunteer(DEFAULT_EMAIL, DEFAULT_PASSWORD);
+        MockHttpServletResponse loginResponse = loginRequest(mvc, new CredentialsVolunteerRequestDto(DEFAULT_EMAIL, DEFAULT_PASSWORD));
+        JwtResponseDto jwtResponseDto = objectMapper.readValue(loginResponse.getContentAsString(), JwtResponseDto.class);
+        Cookie xsrfCookie = loginResponse.getCookie("XSRF-TOKEN");
```
ibaiul, line 116: We don't need to parse the cookie manually with the getCsrfTokenFromCookieHeader, we can get the value properly from the response using Cookie objects. Anyway the parsing was wrong, wouldn't have work when the cookie is not present (NPE) or when the server returns multiple cookies

ibaiul (dismissed):

raulpadilladelgado (approved):
OK

PR #38, opened 2020-06-30
Add email confirmation expiration checker

ManuPerezDev:
Implemented email confirmation expiration checker.


backend/src/test/java/com/huellapositiva/api/EmailAddressConfirmationControllerShould.java
```diff
@@ -73,6 +73,20 @@ void resend_email_should_return_204() throws Exception {
...
                 .andExpect(status().isNoContent());
     }
+
+    @Test
+    void expired_email_should_return_410() throws Exception{
+        // GIVEN
+        UUID token = UUID.randomUUID();
+        testData.createCredential("email@huellapositiva.com", token, "password", VOLUNTEER_NOT_CONFIRMED);
+
+        Thread.sleep(1000);
```
ibaiul: Use await() instead of sleep

backend/src/main/java/com/huellapositiva/domain/actions/EmailConfirmationAction.java
```diff
@@ -32,9 +32,21 @@
...
 
+    @Value("${huellapositiva.email-confirmation.expiration-time}")
+    private Integer emailExpirationTime;
+
     public void execute(UUID hash) {
         EmailConfirmation emailConfirmation = jpaEmailConfirmationRepository.findByHash(hash.toString())
-                .orElseThrow(() -> new EmailConfirmationHashNotFound("Hash " + hash + " not found"));
+                .orElseThrow(() -> new EmailConfirmationHashNotFound("Hash " + hash + " not found."));
+
+        Calendar expirationDate = Calendar.getInstance();
```
ibaiul: Do not use Calendar, use the newer java.time API -> Instant

backend/src/main/java/com/huellapositiva/application/controller/EmailConfirmationApiController.java
```diff
@@ -24,7 +26,11 @@
     @GetMapping("/{hash}")
     @ResponseStatus(HttpStatus.NO_CONTENT)
     public void confirmEmail(@PathVariable UUID hash) {
-        emailConfirmationAction.execute(hash);
+        try {
+            emailConfirmationAction.execute(hash);
+        }catch (EmailConfirmationExpired e){
```
ibaiul: Don't need to catch the exception. It is already annotated with @ResponseStatus(HttpStatus.GONE)

backend/src/main/java/com/huellapositiva/domain/actions/EmailConfirmationAction.java
```diff
@@ -32,9 +32,21 @@
 
     private final JwtService jwtService;
 
+    @Value("${huellapositiva.email-confirmation.expiration-time}")
+    private Integer emailExpirationTime;
```
ibaiul: use long

backend/src/test/java/com/huellapositiva/util/TestData.java
```diff
@@ -55,6 +56,7 @@ public void resetData() {
     private EmailConfirmation createEmailConfirmation(UUID token) {
         EmailConfirmation emailConfirmation = EmailConfirmation.builder()
                 .email(DEFAULT_EMAIL)
+                .createdOn(new Date())
```
ibaiul: Hibernate will override that so no need to set it

backend/src/test/java/com/huellapositiva/integration/ResendEmailConfirmationActionShould.java
```diff
@@ -71,4 +72,25 @@ void update_hash_and_resend_email() {
         assertThat(initialHash, is(not(newHash)));
         verify(communicationService).sendRegistrationConfirmationEmail(any());
     }
+
+    @Test
+    void update_creation_date() throws Exception{
```
ibaiul: This test should not exist, you can extend a test which already does resendEmailConfirmationAction.execute(); and assert the hash and the updatedOn time were actually updated

backend/src/main/java/com/huellapositiva/infrastructure/orm/repository/JpaEmailConfirmationRepository.java
```diff
@@ -21,6 +22,6 @@
 
     @Modifying
     @Transactional
-    @Query("UPDATE EmailConfirmation ec SET ec.hash = :hash WHERE ec.email = :email")
-    Integer updateHashByEmail(@Param("email") String email, @Param("hash") String hash);
+    @Query("UPDATE EmailConfirmation ec SET ec.hash = :hash, ec.createdOn = :createdOn WHERE ec.email = :email")
```
ibaiul: ~~Hibernate should take care of updating the updatedOn column every time you make changes to the entity~~
You probably need to set the updatedOn time in the updatedOn column expicitely since this query will not be executed by Hibernate

backend/src/main/java/com/huellapositiva/domain/actions/ResendEmailConfirmationAction.java
```diff
@@ -38,7 +40,7 @@ public void execute() {
 
         boolean isEmailNotConfirmed = !credential.getEmailConfirmed();
         if (isEmailNotConfirmed) {
-            Integer updateOperation = jpaEmailConfirmationRepository.updateHashByEmail(email, Token.createToken().toString());
+            Integer updateOperation = jpaEmailConfirmationRepository.updateHashByEmail(email, Token.createToken().toString(), new Date());
```
ibaiul: No need to add the date, hibernate should take care.
You need to create a new column in the EmailConfirmation entity called updateOn and annotate it with @UpdateTimestamp so that hibernate automatically changes the date when the entity is updated . Additionally you will need to declare that column in a new incremental flyway script

ibaiul (changes requested):

backend/src/main/java/com/huellapositiva/domain/actions/ResendEmailConfirmationAction.java
Nachocss: It can also be fixed by changing the type of createdOn from _java.util.Date_ to _java.time.Instant_ and using the annotation **@PrePersist** to set its value before inserting the tuple on the DB and **@UpdateTimestamp** to let Hibernate update its value. This way, there is no need to add more columns in the table and/or create a flyway script (The _Instant_ - _Timestamp_ conversion is supported by default)

Nachocss:
It seems to be a conflict with localstack. Currently updating my OS to check it out

backend/src/main/java/com/huellapositiva/application/controller/EmailConfirmationApiController.java
```diff
@@ -24,7 +24,7 @@
     @GetMapping("/{hash}")
     @ResponseStatus(HttpStatus.NO_CONTENT)
     public void confirmEmail(@PathVariable UUID hash) {
-        emailConfirmationAction.execute(hash);
+            emailConfirmationAction.execute(hash);
```
ibaiul: Remove extra space

backend/src/main/java/com/huellapositiva/infrastructure/orm/model/EmailConfirmation.java
```diff
@@ -27,10 +29,15 @@
     @Column(name = "hash", unique = true)
     private String hash;
 
-    @CreationTimestamp
+    @UpdateTimestamp
     @Column(name = "created_on")
-    private Date createdOn;
+    private Instant createdOn;
```
ibaiul: Leave this column as it was. The idea is to have another column called updatedOn. CreatedOn date is only set when the entity is created in the database and should never ever be overriden again. To track when an entity is changed you have the column Date updatedOn. Do not forget to create the incremental flyway script that add the new column to the table

backend/src/main/java/com/huellapositiva/infrastructure/orm/model/EmailConfirmation.java
```diff
@@ -27,10 +29,15 @@
...
 
     @OneToOne(mappedBy = "emailConfirmation")
     private Credential credential;
+
+    @PrePersist
```
ibaiul: Remove this block, it is done by Hibernate automatically

backend/src/test/java/com/huellapositiva/api/JwtControllerShould.java
```diff
@@ -160,14 +158,15 @@ void support_multiple_sessions() throws Exception {
...
         JwtResponseDto sessionOneJwtDto = objectMapper.readValue(loginResponse.getContentAsString(), JwtResponseDto.class);
-        Thread.sleep(1100);
 
         //WHEN
         // Login with second device
         loginRequest(mvc, loginDto);
         //THEN
         // Access token from first login has been revoked due to the second login
-        assertThrows(InvalidJwtTokenException.class, () -> jwtService.getUserDetails(sessionOneJwtDto.getAccessToken()));
+        await().atMost(1, SECONDS).untilAsserted(() ->
```
ibaiul, line 169: Is this really necessary?

backend/src/test/java/com/huellapositiva/integration/ResendEmailConfirmationActionShould.java
```diff
@@ -68,7 +72,11 @@ void update_hash_and_resend_email() {
         EmailConfirmation newEmailConfirmation = jpaEmailConfirmationRepository.findByEmail(DEFAULT_EMAIL)
                 .orElseThrow(() -> new UsernameNotFoundException("User with username: " + DEFAULT_EMAIL + " was not found."));
         String newHash = newEmailConfirmation.getHash();
-        assertThat(initialHash, is(not(newHash)));
+        Instant newCreationTimestamp = newEmailConfirmation.getCreatedOn();
```
ibaiul: Creation date should never change, we should have a different field called updatedOn

ibaiul (changes requested):

backend/src/main/java/com/huellapositiva/domain/actions/ResendEmailConfirmationAction.java
ibaiul: We do want more columns. CreatedOn clearly declares an intention which is a ONE TIME date where an entity was created. For updates another column is used normally. Similarly, a person has only one date of birth. You cannot have two dates of birth or your date of birth cannot be overriden (you cannot born twice) the same way an entity cannot be created twice.

backend/src/test/java/com/huellapositiva/api/JwtControllerShould.java
Nachocss: Yes, because the token's most precise unit to check the expiration time are _seconds_, so after a jwt revocation, the old tokens are still valid until 1 second has passed, or rather the unit of the timestamp that refers to _seconds_ has changed.
If we, for instance, set **await().atMost(300, MILLISECONDS)** it will sometimes pass the tests and sometimes it won't so, setting the maximum wait at 1 second, we ensure the test will always pass

backend/src/test/java/com/huellapositiva/integration/ResendEmailConfirmationActionShould.java
```diff
@@ -53,10 +56,11 @@ void beforeEach() {
     }
 
     @Test
-    void update_hash_and_resend_email() {
+    void update_hash_and_creation_timestamp_and_resend_email() {
```
ibaiul: Replace 'creation' word from test name

ibaiul (changes requested):
The PR is correct on implementing the expiration date. I just realised that the ResendEmailConfirmationAction functionality is missing a use case. When you try to resend a confirmation email but the email is already validated it should return 409 conflict instead of returning 200 and not sending any email to the user. With 409 we are giving the frontend a clue to display a message that the email is already validated. I would write a test for that case as well. ResendEmailConfirmationAction.java -> line 40.
You can argue that fixing that should be a different task, then I will approve straight away.

Nachocss:
Should EmailConfirmationAction check that the email is not confirmed yet?
`boolean isEmailConfirmed = emailConfirmation.getCredential().getEmailConfirmed(); `
`if (isEmailConfirmed) throw new EmailConfirmationAlreadyConfirmed("Email is already confirmed");`

Nachocss:
Since I would need to recreate the changes on Security made today in [PR 39](https://github.com/ayudadigital/huelladigital-backend/pull/39) in order to use our new test library _RestAssured_, I consider it will be better to stick with MockMvc for the last use case and then resolve conflicts once one of the PR is already merged

ibaiul (dismissed):

ibaiul:
> Since I would need to recreate the changes on Security made today in [PR 39](https://github.com/ayudadigital/huelladigital-backend/pull/39) in order to use our new test library _RestAssured_, I consider it will be better to stick with MockMvc for the last use case and then resolve conflicts once one of the PR is already merged

Agree

ibaiul:
> Should EmailConfirmationAction check that the email is not confirmed yet?
> `boolean isEmailConfirmed = emailConfirmation.getCredential().getEmailConfirmed(); `
> `if (isEmailConfirmed) throw new EmailConfirmationAlreadyConfirmed("Email is already confirmed");`

Yes, definitely

RubenZagon:
> Since I would need to recreate the changes on Security made today in [PR 39](https://github.com/ayudadigital/huelladigital-backend/pull/39) in order to use our new test library _RestAssured_, I consider it will be better to stick with MockMvc for the last use case and then resolve conflicts once one of the PR is already merged

The [PR 39](https://github.com/ayudadigital/huelladigital-backend/pull/39) its ready to merge. I will finish the documentation with anothers PR's more specifics with the feature

Nachocss (approved):

PR #41, opened 2020-07-16
Feature/organizations and proposals

ibaiul:


backend/src/main/java/com/huellapositiva/application/controller/OrganizationApiController.java
```diff
@@ -0,0 +1,64 @@
...
+                            content = @Content()
+                    )
+            }
+    )
+    @PostMapping("/register")
+    @ResponseBody
+    public void registerOrganization(@RequestBody OrganizationRequestDto dto, HttpServletRequest req) {
+        try {
+            String authHeader = req.getHeader(HttpHeaders.AUTHORIZATION);
+            String employeeEmail = jwtService.getUserDetails(authHeader.replace("Bearer ", "")).getFirst();
```
ibaiul: Don't need this, you can inject Principals, Authentication, or other similar the same way HttpServletRequest has been injected

backend/src/main/java/com/huellapositiva/application/controller/OrganizationApiController.java
```diff
@@ -0,0 +1,64 @@
...
+                            description = "Ok, organization registered successful"
+                    ),
+                    @ApiResponse(
+                            responseCode = "409",
+                            description = "Conflict, could not register the organization.",
+                            content = @Content()
+                    )
+            }
+    )
+    @PostMapping("/register")
```
ibaiul: The convention for REST APIs states that the root path is used to create a resource; this means that path should be POST /organizations, therefore @PostMapping only

backend/src/main/java/com/huellapositiva/application/controller/OrganizationApiController.java
```diff
@@ -0,0 +1,64 @@
...
+    @ResponseBody
+    public void registerOrganization(@RequestBody OrganizationRequestDto dto, HttpServletRequest req) {
+        try {
+            String authHeader = req.getHeader(HttpHeaders.AUTHORIZATION);
+            String employeeEmail = jwtService.getUserDetails(authHeader.replace("Bearer ", "")).getFirst();
+            OrganizationEmployee organizationEmployee = jpaOrganizationEmployeeRepository.findByEmail(employeeEmail)
+                    .orElseThrow( () -> new RuntimeException("Could not retrieve the organization employee by his email."));
+            registerOrganizationAction.execute(dto, organizationEmployee);
+        } catch (Exception ex) {
+            throw new ResponseStatusException(HttpStatus.CONFLICT, "Could not register the user");
```
ibaiul: Any exception should not lead to a 409 conflict, only the DataIntegrityViolationExceptions should.

backend/src/main/java/com/huellapositiva/application/controller/OrganizationApiController.java
```diff
@@ -0,0 +1,64 @@
...
+            }
+    )
+    @PostMapping("/register")
+    @ResponseBody
+    public void registerOrganization(@RequestBody OrganizationRequestDto dto, HttpServletRequest req) {
+        try {
+            String authHeader = req.getHeader(HttpHeaders.AUTHORIZATION);
+            String employeeEmail = jwtService.getUserDetails(authHeader.replace("Bearer ", "")).getFirst();
+            OrganizationEmployee organizationEmployee = jpaOrganizationEmployeeRepository.findByEmail(employeeEmail)
+                    .orElseThrow( () -> new RuntimeException("Could not retrieve the organization employee by his email."));
```
ibaiul: This exception should be mapped to a 403, not 409

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -0,0 +1,153 @@
...
+                            content = @Content()
+                    ),
+                    @ApiResponse(
+                            responseCode = "409",
+                            description = "Conflict, could not register proposal.",
+                            content = @Content()
+                    )
+            }
+    )
+    @PostMapping("/register")
```
ibaiul: To create a resource the endpoint should be POST /api/v1/proposals

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -0,0 +1,153 @@
...
+    @RolesAllowed({"ORGANIZATION_EMPLOYEE","ORGANIZATION_EMPLOYEE_NOT_CONFIRMED"})
+    @ResponseBody
+    public void createProposal(@RequestBody ProposalRequestDto dto, HttpServletRequest req) {
+        try {
+            registerProposalAction.execute(dto, req);
+        } catch (UserNotConfirmed e){
+            throw new ResponseStatusException(HttpStatus.UNPROCESSABLE_ENTITY, "Employee is not confirmed yet");
```
ibaiul: This should never happen if you restrict the action to ORGANIZATION_EMPLOYEE role only

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -0,0 +1,153 @@
...
+                    ),
+                    @ApiResponse(
+                            responseCode = "409",
+                            description = "Conflict, could not register proposal.",
+                            content = @Content()
+                    )
+            }
+    )
+    @PostMapping("/register")
+    @RolesAllowed({"ORGANIZATION_EMPLOYEE","ORGANIZATION_EMPLOYEE_NOT_CONFIRMED"})
```
ibaiul: NOT_CONFIRMED role should probably not be allow to perform this action.

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -0,0 +1,153 @@
...
+    @ResponseBody
+    public void createProposal(@RequestBody ProposalRequestDto dto, HttpServletRequest req) {
+        try {
+            registerProposalAction.execute(dto, req);
+        } catch (UserNotConfirmed e){
+            throw new ResponseStatusException(HttpStatus.UNPROCESSABLE_ENTITY, "Employee is not confirmed yet");
+        }catch (Exception ex) {
+            throw new ResponseStatusException(HttpStatus.CONFLICT, "Could not register the proposal");
```
ibaiul: If we are not able to determine the error beforehand this should probably be 500 for unknown errors, otherwise catch the errors you expect

backend/src/main/java/com/huellapositiva/application/exception/UserNotConfirmed.java
```diff
@@ -0,0 +1,11 @@
+package com.huellapositiva.application.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(HttpStatus.PRECONDITION_FAILED)
```
ibaiul: This exception should not be necessary if you restrict actions to the roles that are CONFIRMED only

backend/src/main/java/com/huellapositiva/infrastructure/orm/model/Credential.java
```diff
@@ -40,4 +38,16 @@
     @OneToOne(optional = false, cascade = CascadeType.ALL)
     @JoinColumn(name = "email_confirmation_id")
     private EmailConfirmation emailConfirmation;
+
+    @Override
```
ibaiul: @ToString annotation allows you to exclude fields and will autogenerate the method for you

backend/src/main/resources/db/migration/V1_0_5__Add_Organization_Employees.sql
```diff
@@ -0,0 +1,18 @@
+DROP TABLE organization CASCADE;
```
ibaiul: Why DROP?

backend/src/main/resources/db/migration/V1_0_5__Add_Organization_Employees.sql
```diff
@@ -0,0 +1,18 @@
...
+
```
ibaiul: This script (V_1_0_5) can be merged into the previous one. Isn't that part of the same feature?

backend/src/main/resources/db/migration/V1_0_5__Add_Organization_Employees.sql
```diff
@@ -0,0 +1,18 @@
...
+CREATE TABLE organization
+(
+    id              SERIAL PRIMARY KEY NOT NULL,
+    name            VARCHAR(255) NOT NULL UNIQUE
+);
+
+CREATE TABLE organization_users(
```
ibaiul: How about organization_member?

backend/src/main/resources/db/migration/V1_0_5__Add_Organization_Employees.sql
```diff
@@ -0,0 +1,18 @@
...
+    id              SERIAL PRIMARY KEY NOT NULL,
+    credential_id   INTEGER NOT NULL,
+    organization_id INTEGER,
+    FOREIGN KEY (credential_id) REFERENCES credentials(id),
+    FOREIGN KEY (organization_id) REFERENCES organization(id)
+);
+
+UPDATE roles SET name = 'ORGANIZATION_EMPLOYEE' WHERE name = 'ORGANIZATION';
```
ibaiul: How about ORGANIZATION_MEMBER?

backend/src/main/resources/db/migration/V1_0_6__Change_Organization_Employees_Table_Name.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE organization_users RENAME TO organization_employees
```
ibaiul: Merge the content into v1_0_4. The same way we deliver one commit (squashed) for the feature, we should deliver only one script for the feature as well so that it is easier to track history and rollback in case of a disaster.

backend/src/main/java/com/huellapositiva/domain/actions/JoinProposalAction.java
```diff
@@ -0,0 +1,23 @@
...
+@RequiredArgsConstructor
+@Service
+public class JoinProposalAction {
+
+    private final ProposalService proposalService;
+
+    private final VolunteerService volunteerService;
+
+    public void execute(Integer proposalId){
+        String volunteerEmail = (String) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
```
ibaiul: Retrieving the volunteerEmail could be done at controller level and passing it as a parameter so that we decouple this action's business logic from Spring. It would be a better OO approach with many benefits including loose decoupling, testability, maintainability, separation of concerns, ...

backend/src/main/resources/db/migration/V1_0_5__Add_Organization_Employees.sql
Nachocss: If I recall correctly, this statement was written because we could not delete the constraint related to this table in a proper way since the constraint ID was generated automatically by the DB and its value could vary between the local docker image and the one that Jenkins builds.
That said, it will be fixed before closing this branch to develop by merging all the flyway's changes into one single script as you said in the other comments.

backend/src/main/resources/db/migration/V1_0_5__Add_Organization_Employees.sql
Nachocss: Yes, it will be merged soon. 
The reason to have multiple scripts is to not have troubles with the flyway schema history, that does not allow to execute the application if there is changes in previous scripts.
However, now I am wondering if Jenkins does not persist the flyway_schema_history table between builds

backend/src/main/resources/db/migration/V1_0_5__Add_Organization_Employees.sql
ibaiul: Ok, understood.
In any case we should try to generate those "constraint ID"s ourselves whenever is possible so that we have full control of any DB aspect in case of needing to write a rollback script. Dropping a table which could potentially hold data stored in it, is not something we can do once we are in a prod environment. Please review which constraints could not be undone so that they are created properly.

backend/src/main/resources/db/migration/V1_0_5__Add_Organization_Employees.sql
ibaiul: Our app uses an in-memory DB while running the tests every build so all the scripts are applied from scratch.
The problem could come only between deploys but since this feature is still under development and has not been merged to develop branch it should be fine. You might face issues though if you run your application locally with the docker-compose POSTGRES image which you can easily fix by removing the volume it uses.

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -0,0 +1,124 @@
...
+                            description = "Ok, proposal register successful"
+                    )
+            }
+    )
+    @PostMapping
+    @RolesAllowed("ORGANIZATION_MEMBER")
+    @ResponseBody
+    @ResponseStatus(HttpStatus.CREATED)
+    public void createProposal(@RequestBody ProposalRequestDto dto) {
+        String employeeEmail = (String) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
```
ibaiul: You can inject the Authentication object as a parameter to the method instead of using static methods of the SecurityContext which are not very maintainable.

backend/src/main/java/com/huellapositiva/application/controller/VolunteerApiController.java
```diff
@@ -58,10 +57,9 @@
                     )
             }
     )
-
-
-    @PostMapping("/register")
+    @PostMapping
     @ResponseBody
+    @ResponseStatus(HttpStatus.CREATED)
```
ibaiul, line 66: To meet REST and HTTP standards if we return 201 the convention states that we should send back the URI ("/volunteer/id") of the created resource in the Location header of the response

Nachocss:
We should merge this branch at this point so the frontend team can start working with the new endpoints.
We added a setter that publish every proposal for this purpose in **ProposalApiController** but shall be removed when an admin account enters into action

backend/src/main/java/com/huellapositiva/application/controller/OrganizationApiController.java
```diff
@@ -0,0 +1,64 @@
...
+                    ),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Internal server error, could not register the organization.",
+                            content = @Content()
+                    )
+            }
+    )
+    @PostMapping
+    @RolesAllowed("ORGANIZATION_MEMBER")
```
ibaiul, line 65: Should this action be performed only by ADMIN role? Otherwise any MEMBER would be able to create unlimited Organizations? Check it is aligned with Product colleagues or Juan from Huellapositiva. This affects the UX.

backend/src/main/java/com/huellapositiva/application/controller/OrganizationApiController.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.web.bind.annotation.*;
+import org.springframework.web.server.ResponseStatusException;
+
+import javax.annotation.security.RolesAllowed;
+
+@RestController
+@AllArgsConstructor
+@Tag(name = "Organizations", description = "The organization API")
+@RequestMapping("/api/v1/organizations")
+@EnableWebSecurity
```
ibaiul: Annotation not required, already present in SecurityConfig.

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -0,0 +1,133 @@
...
+import javax.annotation.security.RolesAllowed;
+import javax.persistence.EntityNotFoundException;
+import javax.servlet.http.HttpServletResponse;
+import java.net.URI;
+
+@RestController
+@AllArgsConstructor
+@Tag(name = "Proposal Service", description = "The proposals API")
+@RequestMapping("/api/v1/proposals")
+@EnableWebSecurity
```
ibaiul: Annotation not required, already present in SecurityConfig.

backend/src/main/java/com/huellapositiva/application/controller/OrganizationMemberApiController.java
```diff
@@ -0,0 +1,89 @@
...
+
+import javax.servlet.http.HttpServletResponse;
+import java.net.URI;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@RestController
+@AllArgsConstructor
+@Tag(name = "Organization Employee", description = "The organization employee API")
+@RequestMapping("/api/v1/organizationmember")
```
ibaiul: What about /api/v1/member?

backend/src/main/java/com/huellapositiva/application/controller/OrganizationMemberApiController.java
```diff
@@ -0,0 +1,89 @@
...
+            Integer id = registerOrganizationMemberAction.execute(dto);
+            String username = dto.getEmail();
+            List<String> roles = jpaRoleRepository.findAllByEmailAddress(username).stream().map(Role::getName).collect(Collectors.toList());
+            URI uri = ServletUriComponentsBuilder.fromCurrentRequest()
+                    .path("/{id}").buildAndExpand(id)
+                    .toUri();
+            res.addHeader(HttpHeaders.LOCATION, uri.toString());
+            return jwtService.create(username, roles);
+        } catch (PasswordNotAllowed pna) {
+            throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Password doesn't meet minimum length");
```
ibaiul: Don't need to catch it, it already maps to 400

backend/src/main/java/com/huellapositiva/application/controller/OrganizationMemberApiController.java
```diff
@@ -0,0 +1,89 @@
...
+        } catch (FailedToPersistUser ex) {
```
ibaiul: This class should be called ConflictPersistingUserException and be mapped to @ResponseStatus(409). This way you do not need to catch the exception.

backend/src/main/java/com/huellapositiva/infrastructure/orm/model/OrganizationMember.java
```diff
@@ -0,0 +1,29 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Integer id;
+
+    @JoinColumn(name = "credential_id")
+    @OneToOne(optional = false, cascade = CascadeType.ALL, fetch = FetchType.EAGER)
+    private Credential credential;
+
+    @JoinColumn(name = "organization_id")
+    @ManyToOne(fetch = FetchType.EAGER)
+    private Organization joinedOrganization;
```
ibaiul, line 28: This relation should be modeled in the other end as well.

backend/src/main/java/com/huellapositiva/infrastructure/orm/model/Proposal.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Column(name = "published")
+    private Boolean published;
+
+    @ManyToMany(fetch = FetchType.EAGER)
+    @JoinTable(name = "volunteers_proposals",
+            joinColumns = {@JoinColumn(name = "proposal_id", referencedColumnName = "id")},
+            inverseJoinColumns = {@JoinColumn(name = "volunteer_id", referencedColumnName = "id")}
+    )
+    private Set<Volunteer> inscribedVolunteers;
```
ibaiul: What about joinedVolunteers, isn't the endpoint /proposals/{id}/join ?

backend/src/test/java/com/huellapositiva/api/OrganizationMemberControllerShould.java
```diff
@@ -0,0 +1,80 @@
...
+    void register_member_and_return_201_and_tokens() throws Exception {
+        // GIVEN
+        CredentialsOrganizationMemberRequestDto dto = new CredentialsOrganizationMemberRequestDto(DEFAULT_EMAIL, DEFAULT_PASSWORD);
+
+        // WHEN
+        MockHttpServletResponse response = mvc.perform(post("/api/v1/organizationmember")
+                .content(objectMapper.writeValueAsString(dto))
+                .contentType(APPLICATION_JSON)
+                .accept(APPLICATION_JSON))
+                .andExpect(status().isCreated())
```
ibaiul, line 66: Location header can be asserted with andExpect()

Nachocss (approved):

PR #43, opened 2020-08-28
Feature/Simple Storage Service (AWS S3) for persisting images and documents

Nachocss:


backend/pom.xml
```diff
@@ -14,6 +14,7 @@
     <properties>
         <assertj.version>3.15.0</assertj.version>
         <aws.ses.version>1.11.789</aws.ses.version>
+        <aws.s3.version>1.11.848</aws.s3.version>
```
ibaiul: try to use the same AWS SDK for all AWS services you integrate to avoid incompatibilities . SES, S3, ...

backend/src/main/java/com/huellapositiva/domain/service/RemoteStorageService.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class RemoteStorageService {
+
+    @Autowired
+    private final StorageService storageService;
+
+    public URL uploadProposalImage(MultipartFile file, String proposalId) throws IOException {
+        String destinationFileName = proposalId + "." + file.getContentType();
```
ibaiul: Since all the images will be publicly available from S3, please make sure the name/route of the images is not guessable otherwise anyone would be able to access all of them.

You can only have one file with the same name within a bucket so it would be advisable to assign every image a random uuid as the name and store that UUID in the database. That approach would improve guessability of files and the flexibility of the platform by allowing multiple images to be bound to the same proposal in the future.

backend/src/main/java/com/huellapositiva/infrastructure/AwsS3Config.java
```diff
@@ -0,0 +1,63 @@
...
+    public AmazonS3 getLocalstackAwsS3Client() {
+        log.info("Localstack Amazon S3 client enabled");
+        BasicAWSCredentials awsCreds = new BasicAWSCredentials(awsS3Properties.getAccessKey(), awsS3Properties.getSecretKey());
+        AmazonS3 s3client = AmazonS3ClientBuilder
+                .standard().withPathStyleAccessEnabled(true)
+                .withCredentials(new AWSStaticCredentialsProvider(awsCreds))
+                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(awsS3Properties.getEndpoint(), awsS3Properties.getRegion()))
+                .build();
+
+        String bucketName = awsS3Properties.getBucketName();
```
ibaiul, line 55: DRY -> duplicated

backend/src/main/java/com/huellapositiva/infrastructure/AwsS3Config.java
```diff
@@ -0,0 +1,63 @@
...
+        if(!s3client.doesBucketExistV2(bucketName)) {
+            s3client.createBucket(new CreateBucketRequest(bucketName));
+        }
+
+        return s3client;
```
ibaiul: DRY -> duplicated

backend/src/main/java/com/huellapositiva/infrastructure/AwsStorageService.java
```diff
@@ -0,0 +1,34 @@
...
+    private AwsS3Properties awsS3Properties;
+
+    @Override
+    public URL upload(File file, String key) {
+        String bucketName = awsS3Properties.getBucketName();
+        awsS3Client.putObject(
+                bucketName,
+                key,
+                file
+        );
```
ibaiul: Doesn't it fit in one line? :-)

backend/src/main/resources/application-dev.properties
```diff
@@ -17,6 +18,12 @@ aws.ses.secretKey=${AWS_SECRET_KEY}
 aws.ses.endpoint=${AWS_SES_ENDPOINT_HOST}
 aws.ses.region=${AWS_REGION}
 
+aws.s3.accessKey=${AWS_S3_ACCESS_KEY_ID}
+aws.s3.secretKey=${AWS_S3_SECRET_KEY}
```
ibaiul: It should use the same credentials as SES.

In the AWS console we need to create an IAM role called "backend" with permissions for both S3 and SES, then create a user with that role and finally use the credentials of that user.

Nachocss:
Why Jenkins does not support MONTHS as a ChronoUnit when adding months to a LocalDate or an Instant?

ManuPerezDev (approved):

PR #45, opened 2020-09-11
Feature/Proposal's revision notification service

Nachocss:


Nachocss:
Let's avoid working on this branch any features that are not strictly related to Proposal's revision notification service. Request a review if you think we should merge it  @mariasoria @juanmaMacGyverCode

mariasoria:
Good morning @ibaiul, how are you?

Could you help us merging this branch, please? Could you have a look at the changes and approve the review?

Thank a million

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -235,7 +248,172 @@ public void createProposalAsReviser(@RequestPart("dto") MultipartFile dtoMultipa
...
+        try {
+            URI uri = ServletUriComponentsBuilder.fromCurrentRequest()
+                    .path(PATH_ID).buildAndExpand(id)
+                    .toUri();
+            dto.setReviserEmail(reviserEmail);
+            submitProposalRevisionAction.execute(id, dto, uri);
+        } catch (NullPointerException ex) {
+            throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Could not parse the revision, due to missing data.");
+        } catch (EntityNotFoundException ex) {
+            throw new ResponseStatusException(HttpStatus.BAD_REQUEST, PROPOSAL_DOESNT_EXIST);
```
manumoranchel: As discussed, return 404. Alter Swagger and tests accordingly

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -235,7 +248,172 @@ public void createProposalAsReviser(@RequestPart("dto") MultipartFile dtoMultipa
...
+    public void submitProposalRevision(@PathVariable String id,
+                                       @RequestBody ProposalRevisionDto dto,
+                                       @AuthenticationPrincipal String reviserEmail) {
+        try {
+            URI uri = ServletUriComponentsBuilder.fromCurrentRequest()
+                    .path(PATH_ID).buildAndExpand(id)
+                    .toUri();
+            dto.setReviserEmail(reviserEmail);
+            submitProposalRevisionAction.execute(id, dto, uri);
+        } catch (NullPointerException ex) {
```
manumoranchel: remember not to catch NPEs

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -125,11 +133,16 @@ public void createProposal(@RequestPart("dto") MultipartFile dtoMultipart,
     )
     @GetMapping("/{id}")
     @ResponseStatus(HttpStatus.OK)
-    public ProposalResponseDto getProposal(@PathVariable String id) {
+    public ProposalResponseDto getProposal(@PathVariable String id, HttpServletResponse res) throws IOException {
         try {
             return fetchProposalAction.execute(id);
-        } catch (EntityNotFoundException | ProposalNotPublished e) {
-            throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Proposal with ID " + id + "does not exist or is not published.");
+        } catch (EntityNotFoundException | ProposalNotPublic e) {
```
manumoranchel: This is not recommended. An API should be agnostic of what client is performing the request. In other words a request can be performed by another application, a web browser, etc. 
Here you are assuming the request is coming from a **browser** and performing a redirect. Such redirect, if needed, should be done from the the **FRONTEND**.

If the request is trying to access an end-point that has been _PERMANTLY_ or _TEMPORARILY_ moved, you can return dedicated _HTTPStatus_ codes. But remember, the client (in huella is the frontend) should perform the redirection

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -235,7 +248,172 @@ public void createProposalAsReviser(@RequestPart("dto") MultipartFile dtoMultipa
...
+                            responseCode = "400",
+                            description = "Bad request, a conflict was encountered while attempting to persist the proposals."
+                    ),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Internal server error, could not fetch the user data due to a connectivity issue."
+                    )
+            }
+    )
+    @RolesAllowed("REVISER")
```
manumoranchel, line 276: (MINOR) Not necesary to change it.
I can see there is class _com.huellapositiva.domain.model.valueobjects.Roles_ This holds an enum with all the roles handled by the application. Ideally, instead of having "magic strings" we should use the available configured ENUMs.
One of the benefits is when a change is needed, you can change it in one place.

backend/src/main/java/com/huellapositiva/application/dto/ProposalLiteDto.java
```diff
@@ -16,6 +16,8 @@
     @NotEmpty
     private final String title;
 
+    private final String status;
```
manumoranchel, line 19: Not for this PullRequest
I can see many common fields in the **Proposal*Dto**. Maybe we could use inheritance relationship to avoid duplicate fields?

backend/src/main/java/com/huellapositiva/application/exception/ProposalNotPublic.java
```diff
@@ -0,0 +1,16 @@
+package com.huellapositiva.application.exception;
+
+/**
+ * This exception is thrown when a proposal is neither PUBLISHED nor FINISHED
+ * @see com.huellapositiva.domain.model.valueobjects.ProposalStatus
+ */
+public class ProposalNotPublic extends RuntimeException {
```
manumoranchel: Not sure what's the agreement in this project since I can see several Exceptions with similar naming.
I usually like to suffix new exceptions with **Exception**, in this case **ProposalNotPublicException** this ease the readiblity and adds context when handling them.
I leave it up to you wether you want to implement it or not

backend/src/main/java/com/huellapositiva/domain/actions/FetchPaginatedProposalsAction.java
```diff
@@ -14,26 +14,57 @@
 
     private final ProposalRepository proposalRepository;
 
+    /**
```
manumoranchel, line 17: I love a good javadoc. Good work

backend/src/main/java/com/huellapositiva/domain/actions/RegisterESALAction.java
```diff
@@ -30,7 +41,18 @@ public void execute(ESALRequestDto dto, EmailAddress loggedContactPersonEmail) {
...
+     *
+     * @param dto contains the info to create a new ESAL
+     */
     public void execute(ESALRequestDto dto) {
-        esalService.create(dto);
+        try {
+            ESAL esal = new ESAL(dto.getName(), esalRepository.newId());
+            esalRepository.saveAsReviser(esal);
+        } catch (DataIntegrityViolationException ex) {
+            log.error("Unable to persist the proposal due to a conflict.", ex);
```
manumoranchel: Good work on adding the logs. In my opinion, you should have a log.error right after catching the exception

backend/src/main/java/com/huellapositiva/domain/actions/SubmitProposalRevisionAction.java
```diff
@@ -0,0 +1,47 @@
...
+    private final EmailCommunicationService communicationService;
+
+    private final ProposalRepository proposalRepository;
+
+    private final ContactPersonRepository contactPersonRepository;
+
+    private final CredentialsRepository credentialsRepository;
+
+    /**
+     * TO BE DEFINED!!
```
manumoranchel: what happened here?????? :-)

backend/src/main/java/com/huellapositiva/domain/model/entities/Reviser.java
```diff
@@ -0,0 +1,19 @@
+package com.huellapositiva.domain.model.entities;
+
+import com.huellapositiva.domain.model.valueobjects.EmailAddress;
+import com.huellapositiva.domain.model.valueobjects.Id;
+
+public class Reviser extends User {
+
+    public Reviser(EmailAddress emailAddress, Id id) {
```
manumoranchel, line 8: can lombok help here?

backend/src/main/java/com/huellapositiva/domain/service/ESALContactPersonService.java
```diff
@@ -25,23 +24,31 @@
...
 
-    public Integer updateJoinedOrganization(JpaContactPerson employee, JpaESAL organization) {
-        return esalContactPersonRepository.updateOrganization(employee.getId(), organization);
-    }
-
-    public Optional<JpaContactPerson> findByEmail(String email){
-        return esalContactPersonRepository.findByEmail(email);
+    /**
+     *
+     * TO BE DEFINED
```
manumoranchel: Shall we define it????

manumoranchel (commented):
Looks good overall, good work. I've only commented the code becasue I have no context of the requirements.
I'll leave @ibaiul to have the final decision :-)

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
juanmaMacGyverCode: Fixed. Test swagger and controller have been modified.

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
juanmaMacGyverCode: Fixed. Removed the catch and managed null value of the feedback.

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
juanmaMacGyverCode: Removed the redirection and sent a dedicated HTTP status code and updated tests

backend/src/main/java/com/huellapositiva/application/exception/ProposalNotPublic.java
juanmaMacGyverCode: @carlosble 

We dont know if it is an agreement not to use the word exception at the end of the name the class exception. Do you think we should change it? Many thanks.

backend/src/main/java/com/huellapositiva/application/dto/ProposalLiteDto.java
juanmaMacGyverCode: We think that this could be done in a future refactor in develop branch

backend/src/main/java/com/huellapositiva/application/exception/ProposalNotPublic.java
ibaiul: It is a Java convention to suffix with xxxException

backend/src/main/java/com/huellapositiva/domain/actions/UploadCurriculumVitaeAction.java
```diff
@@ -20,14 +19,21 @@
...
+     * @param cv
+     * @param volunteerEmail
+     * @throws IOException when the cv is corrupted
+     */
     public void execute(MultipartFile cv, String volunteerEmail) throws IOException {
-        Volunteer volunteer = volunteerService.findVolunteerByEmail(volunteerEmail);
-        URL cvUrl = remoteStorageService.uploadVolunteerCV(cv, volunteer.getId().toString());
-        volunteer.setCurriculumVitae(cvUrl);
-        volunteerRepository.updateCurriculumVitae(volunteer);
+        if (cv != null) {
```
ibaiul: If cv is null this action does nothing and succeeds. If I want to upload a cv and there is no cv, I would rather fail the request with a 400

backend/src/main/java/com/huellapositiva/domain/model/entities/User.java
```diff
@@ -2,17 +2,52 @@
...
+        this.passwordHash = passwordHash;
+        this.id = id;
+    }
+
+    public String getFullName() {
+        if (name == null) {
+            return null;
+        }
+        if (surname == null) {
+            return String.format("%s", name);
```
ibaiul: Why not return variable name instead of formatting?

backend/src/main/java/com/huellapositiva/domain/model/entities/User.java
```diff
@@ -2,17 +2,52 @@
...
+        this.emailAddress = emailAddress;
+        this.passwordHash = passwordHash;
+        this.id = id;
+    }
+
+    public String getFullName() {
+        if (name == null) {
+            return null;
+        }
+        if (surname == null) {
```
ibaiul: else if would work too instead of having multiple ifs

backend/src/main/java/com/huellapositiva/domain/model/valueobjects/ProposalRevisionEmail.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public boolean hasFeedback(){
+        return hasFeedback;
+    }
+
+    public String getProposalURL() {
+        try {
+            return this.proposalURI.toURL().toExternalForm();
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
```
ibaiul, line 47: Logs can help to debug potential issues

backend/src/main/java/com/huellapositiva/infrastructure/orm/repository/JpaStatusRepository.java
```diff
@@ -0,0 +1,17 @@
...
+import com.huellapositiva.infrastructure.orm.entities.JpaStatus;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+
+@Repository
+public interface JpaStatusRepository extends JpaRepository<JpaStatus, Integer> {
```
ibaiul: StatusRepository is very ambiguous. What status are we referring to? I would call it JpaProposalStatus otherwise anyone can think it is the status of a volunteer, esal or any other kind of thing.

The same applies to the JpaStatus class

backend/src/main/resources/db/migration/V1_0_7__Add_Users_Personal_Information.sql
```diff
@@ -0,0 +1,28 @@
+ALTER TABLE credentials
+ADD COLUMN name VARCHAR(255);
+
+ALTER TABLE credentials
+ADD COLUMN surname VARCHAR(255);
+
+ALTER TABLE proposals
+DROP COLUMN published;
```
ibaiul, line 8: This is fine for now but once we are live we cannot remove a column this way. It would not allow us to rollback the deployment and would cause complete data corruption. I can explain in a future meeting if interested.

backend/src/main/resources/db/migration/V1_0_7__Add_Users_Personal_Information.sql
```diff
@@ -0,0 +1,28 @@
...
+
+ALTER TABLE proposals
+ADD COLUMN status_id INTEGER;
```
ibaiul: If this column is mandatory we need to set a value (default or based on the previous "published" column) to all the existing proposals in the database or the data can get corrupted. It is not that important now since we can delete the DB schema, but it won't be possible in the future once we are live

backend/src/main/java/com/huellapositiva/domain/actions/SubmitProposalRevisionAction.java
mariasoria: Ooops, we forgot to add the javadoc.
We've just done it ;)

backend/src/main/java/com/huellapositiva/domain/service/ESALContactPersonService.java
mariasoria: Yes! DONE ;)

backend/src/main/resources/db/migration/V1_0_7__Add_Users_Personal_Information.sql
manumoranchel: I understand the reasoning not to drop columns but I would love to hear your experience in rolling back databases to a previous state. It was a challenge in the previous projects I worked and I will always be interested in success stories :-)

backend/src/main/java/com/huellapositiva/domain/actions/UploadCurriculumVitaeAction.java
mariasoria: Changed. Now in this action file we are checking if the file comes empty or if it is not a file attached. In that case, we fail the request with a 400 error.

backend/src/main/java/com/huellapositiva/application/exception/ProposalNotPublic.java
mariasoria: We have just renamed all those exceptions, adding the suffix.

backend/src/main/java/com/huellapositiva/domain/model/entities/User.java
mariasoria: Changed ;)

backend/src/main/java/com/huellapositiva/domain/model/entities/User.java
mariasoria: Done!

backend/src/main/java/com/huellapositiva/infrastructure/orm/repository/JpaStatusRepository.java
mariasoria: Yes, we agree. Changed applied.

backend/src/main/resources/db/migration/V1_0_7__Add_Users_Personal_Information.sql
mariasoria: Yes, we would like you to explain us this further, please 🙏

backend/src/main/resources/db/migration/V1_0_7__Add_Users_Personal_Information.sql
mariasoria: We have set the status by default to be REVIEW_PENDING.

backend/src/main/java/com/huellapositiva/domain/model/entities/Reviser.java
mariasoria: Hi Manu, what do you mean? 🎵

manumoranchel: From a code perspective, the code looks good as it is. I have no argument against it. However Lombok can support with certain common operations like the constructors.
If you want to play with it, I'd look at @NoArgsConstructor, @AllArgsConstructor annotations and see if you can support on them to make the code look nicer.

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -96,7 +97,7 @@ public void createProposal(@RequestPart("dto") MultipartFile dtoMultipart,
                                @AuthenticationPrincipal String contactPersonEmail,
                                HttpServletResponse res) throws IOException {
         ProposalRequestDto dto = objectMapper.readValue(dtoMultipart.getBytes(), ProposalRequestDto.class);
-        dto.setStatus(PUBLISHED.getId());
```
ibaiul: If the status cannot be set by the user, I would remove the status field from the DTO class (request specification)

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -96,7 +97,7 @@ public void createProposal(@RequestPart("dto") MultipartFile dtoMultipart,
...
+        dto.setStatus(REVIEW_PENDING.getId()); // Setting the status when a proposal is created the first time.
```
ibaiul: Setting the default proposal status seems to be responsibility of the register proposal action rather than the controller

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
mariasoria: Hi Ibai, you are right. The status is never set by the user. But, depending on the person who creates the proposal, the status is different.
If the “Reviser” creates it, then, the status is directly “PUBLISHED”, if it is the “ContactPerson”, the status is “REVIEW_PENDING” instead.

So, we have different options.

1st option:
If we remove that attribute from the dto object class, still we need to set that value before saving that info into the DB., which can be done in the action class. But, in order to do that (as goes first to the parser to create the object “proposal” which will be the one we will save in the DB), then, we’ll need to add that argument to the parser, which means, sending 3 parameters to it.

2nd option:
If we don’t remove it, even though is actually not part of a dto, we don’t need to do all that from the 1st option.

What do you think? Better remove it from the dto (knowing that that will mean passing more parameters to the parser) or leave it as it is (knowing, that is not actually part of the dto)??

Thank youuu

PS: We will use that “STATUS” soon in other similar situations when the proposal needs to be edited, for example. But that part is not implemented yet.

backend/src/main/java/com/huellapositiva/domain/model/entities/Reviser.java
mariasoria: Hi Manu! Thank you for your comment. 
We had a look at it, and it seems we can't use those annotations, because those constructors extend those in USER.java. Or do you think we can still use them?? Thanks for your help :)

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
ibaiul: The request DTO should reflect the actual contract of the API, so you should not have an status field and therefore should not deserialise it with the object mapper. I guess that is aligned with your first option.
The internal implementation is up to you, the following are just available options for the action's public contract with the controller.
1- you can define two arguments in the execute method of the action, ProposalRequestDto dto, ProposalStatus status
2- you can create two public methods in the action executeByVolunteer(ProposalRequestDto) and executeByAdmin(ProposalRequestDto dto)
3- you can create two actions that extend an abstract action or implement an abstract action interface

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
mariasoria: OK, yes, you are right. We have just changed it. Thanks!

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
mariasoria: Hi Ibai, we think we've got it.
- We removed from the DTO the status (which makes sense). 
- Renamed the "execute" methods, to "executeByContactPerson" and "executeByAdmin".
- Set the status in the action class right after parsing the DTO into a Proposal object. (Not sure if it is the best option) Could you confirm?

Thank you

backend/src/main/java/com/huellapositiva/domain/model/entities/Reviser.java
manumoranchel: I think you are right. Looking at the documentation in: https://projectlombok.org/features/constructor One can see the different configurations that can be added to the mentioned annotations.
Some lombok annotations allow the flag `callSuper = true` However this feature seems not available to build the Object :-)
It is good as it is. Today we've lean something new.

ibaiul (dismissed):
Next time we should ensure we cover %80 of the new code written.

ibaiul (approved):

mariasoria:
We have merged this branch into develop because the SonarCloud analysis is checking only develop branch, which means that is not even considering the changes done in this branch. After merging, we'll check the analysis again, and will try to solve the issues.

PR #55, opened 2020-11-06
Feature/cancel proposal

mariasoria:


backend/src/main/java/com/huellapositiva/domain/actions/CancelProposalAction.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Service
+public class CancelProposalAction {
+
+    @Autowired
+    ProposalRepository proposalRepository;
+
+    public void executeByReviser(String id){
+        JpaProposalStatus jpaProposalStatus = JpaProposalStatus.builder()
+                .id(6)
```
manumoranchel: What does 6 means????? :-)
This is an example of magic number. A developer reading this code won't be able to know the meaning of the id unless he/she navigates to the ProposalStatus class.
You can set a constant in the CancelProposalAction class but even better use the ProposalStatus.CANCELLED

backend/src/test/java/com/huellapositiva/api/ProposalControllerShould.java
```diff
@@ -701,4 +701,45 @@ void return_404_when_proposal_with_volunteers_not_found() throws Exception{
...
+        testData.createCredential("revisor@huellapositiva.com", UUID.randomUUID(), DEFAULT_PASSWORD, Roles.REVISER);
+        JwtResponseDto jwtResponseDto = loginAndGetJwtTokens(mvc, "revisor@huellapositiva.com", DEFAULT_PASSWORD);
+
+        //WHEN + THEN
+        MockHttpServletResponse fetchResponse = mvc.perform(post(FETCH_PROPOSAL_URI + proposalId + "/cancel")
+                .header(HttpHeaders.AUTHORIZATION, "Bearer " + jwtResponseDto.getAccessToken())
+                .with(csrf())
+                .contentType(MediaType.APPLICATION_JSON)
+                .accept(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
```
manumoranchel: Have you considered checking the value in the database is updated????

backend/src/main/java/com/huellapositiva/domain/actions/CancelProposalAction.java
mariasoria: Yes, that's right. We have just changed it. Thanks!

manumoranchel: did you push the changes? I can't see them :-)

backend/src/test/java/com/huellapositiva/api/ProposalControllerShould.java
mariasoria: Done ;)

backend/src/test/java/com/huellapositiva/api/ProposalControllerShould.java
```diff
@@ -701,4 +700,53 @@ void return_404_when_proposal_with_volunteers_not_found() throws Exception{
...
+
+        MockHttpServletResponse fetchResponseAllProposals = mvc.perform(get(FETCH_PROPOSAL_URI + "0/5/reviser")
+                .header(HttpHeaders.AUTHORIZATION, "Bearer " + jwtResponseDto.getAccessToken())
+                .with(csrf())
+                .contentType(MediaType.APPLICATION_JSON)
+                .accept(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
+                .andReturn().getResponse();
+
+        ListedProposalsDto listedProposalsDto = objectMapper.readValue(fetchResponseAllProposals.getContentAsString(), ListedProposalsDto.class);
```
manumoranchel: I would have done it this way (WARNING: pseudocode) ;-)
```
JpaProposal jpaProposal = jpaProposalRepository.findById(proposalId);
assertThat(jpaProposal.getStatus().contains(CANCELLED.toString())).isTrue();
```
But maybe I am missing something. PR looks good to me

manumoranchel (dismissed):

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -395,11 +399,51 @@ public void submitProposalRevision(@PathVariable String id,
...
+                    ),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Internal server error, could not fetch the user data due to a connectivity issue."
+                    )
+            }
+    )
+    @PostMapping("/{id}/cancel")
+    @RolesAllowed("REVISER")
+    @ResponseStatus(HttpStatus.OK)
```
ibaiul: If we return a successful response without body the code should be 204 NO_CONTENT

backend/src/main/java/com/huellapositiva/domain/actions/CancelProposalAction.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * This method changes the specified proposal status to CANCELLED.
+     *
+     * @param id Proposal id
+     *
+     * (method "updateProposalStatus" returns 0 in case it doesn't find any proposal with that id)
+     */
+    public void executeByReviser(String id) {
+        JpaProposalStatus jpaProposalStatus = JpaProposalStatus.builder()
+                .id(ProposalStatus.CANCELLED.getId())
```
ibaiul, line 26: I wouldn't rely on the hardcoded id's from an enum, they could be misaligned with the entries on the DB. You should use the JpaProposalStatusRepository to retrieve the correct ID.

Additionally, since the IDs of the statuses are static (they do not change) we could even cache the response from the DB to avoid making a DB query everytime we need to retrieve the ids of the statuses

manumoranchel: I suggested the change so I am guilty for this one :-)
IMHO we shouldn't let the database (infrastructure) drive model identifiers but the other way around. Because this class is in the Model, it should talk "model language" and it's identifiers.
If the values differ in the database, the adapter subscribed to that port should be the one performing the transformation.
What am I not understanding here???

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
juanmaMacGyverCode: Updated.

backend/src/main/java/com/huellapositiva/domain/actions/CancelProposalAction.java
mariasoria: Hello guys, 
in this case, the hardcoded enum will be always updated with the DB because that table is every time created from the enum, and not modified at all during the execution, so there won't be any problem with misaligned information.
So far, in the code we are not accessing the DB every time we need to check the id of a specific proposalStatus,  if we do that, then, we should think of what you guys are saying, and decide if we want to query the DB every time or we want to cache the response and access to it when that information is needed.

What do you think is better in this case? Should we cache the proposalStatus table from the DB and update the code to access to it every time needed or leave it like that and trust the hardcoded enum (which unless there's any change, will always match with the info from the DB)?

ibaiul: @manumoranchel I agree with you. The thing is that the ID of the ProposalStatus domain model could already be the enum String value itself and don't require any additional integers at all in your domain model. An integer is something meaningful for a database more than a human in many cases, so sometimes it might not add much value to the domain semantics if used as a hardcoded ID.

manumoranchel: oh I see what you mean. Thanks for the explanation :-)
I was talking to Maria about the change via chat. How about we merge the PR with the changes and we add a technical debt User Story to refactor this functionality keeping the Domain objects numberless and adapting when accessing the database?

ibaiul: Sure.

ibaiul (approved):

PR #56, opened 2020-11-16
Feature/edit password from profile

juanmaMacGyverCode:
Good morning.

We have created a feature to allow the user to change their password from their profile.

Regards.

### DIFF
--- a/backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
+++ b/backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
@@ -1,7 +1,7 @@
 package com.huellapositiva.application.controller;
 
 import com.huellapositiva.application.exception.UserNotFoundException;
-import com.huellapositiva.domain.actions.FetchCredentialsAction;
+import com.huellapositiva.domain.actions.UpdatePasswordAction;
 import com.huellapositiva.domain.exception.InvalidNewPasswordException;
 import com.huellapositiva.domain.exception.NonMatchingPasswordException;
 import io.swagger.v3.oas.annotations.Operation;
@@ -22,7 +22,7 @@ import javax.annotation.security.RolesAllowed;
 public class HandlerPasswordApiController {
 
     @Autowired
-    FetchCredentialsAction credentialsAction;
+    UpdatePasswordAction credentialsAction;
 
     @Operation(
             summary = "Send an email to recovery password",
--- a/backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
+++ b/backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
@@ -14,11 +14,9 @@ import org.springframework.http.HttpStatus;
 import org.springframework.security.core.annotation.AuthenticationPrincipal;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.*;
-import org.springframework.web.multipart.MultipartFile;
 import org.springframework.web.server.ResponseStatusException;
 
 import javax.annotation.security.RolesAllowed;
-import java.io.IOException;
 
 
 @Controller
@@ -27,8 +25,6 @@ public class HandlerPasswordApiController {
 
     @Autowired
     UpdatePasswordAction credentialsAction;
-
-    private final ObjectMapper objectMapper = new ObjectMapper();
 
     @Operation(
             summary = "Send an email to recovery password",
@@ -112,9 +108,8 @@ public class HandlerPasswordApiController {
     @RolesAllowed({"VOLUNTEER", "CONTACT_PERSON"})
     @PostMapping("/editPassword")
     @ResponseStatus(HttpStatus.NO_CONTENT)
-    public void editProfilePassword(@RequestPart("dto") MultipartFile dtoMultipart,
-                                    @AuthenticationPrincipal String email) throws IOException {
-        ChangePasswordDto dto = objectMapper.readValue(dtoMultipart.getBytes(), ChangePasswordDto.class);
+    public void editProfilePassword(@RequestBody(required = true) ChangePasswordDto dto,
+                                    @AuthenticationPrincipal String email) {
         try {
             credentialsAction.executeUpdatePassword(dto, email);
         } catch (NonMatchingPasswordException | InvalidNewPasswordException e) {
--- a/backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
+++ b/backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
@@ -13,8 +13,8 @@ import org.springframework.web.server.ResponseStatusException;
 
 
 @Controller
-@RequestMapping("/api/v1/restore-password")
-public class RecoveryPasswordApiController {
+@RequestMapping("/api/v1/handling-password")
+public class HandlerPasswordApiController {
 
     @Autowired
     FetchCredentialsAction credentialsAction;
--- a/backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
+++ b/backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
@@ -2,14 +2,19 @@ package com.huellapositiva.application.controller;
 
 import com.huellapositiva.application.exception.UserNotFoundException;
 import com.huellapositiva.domain.actions.FetchCredentialsAction;
+import com.huellapositiva.domain.exception.InvalidNewPasswordException;
+import com.huellapositiva.domain.exception.NonMatchingPasswordException;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
+import org.springframework.security.core.annotation.AuthenticationPrincipal;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.server.ResponseStatusException;
+
+import javax.annotation.security.RolesAllowed;
 
 
 @Controller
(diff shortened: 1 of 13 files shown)


backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
```diff
@@ -76,4 +81,38 @@ public void sendEmailRecovery(@RequestParam("email") String email){
...
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Internal server error, could not fetch the user data due to a connectivity issue."
+                    )
+            }
+    )
+    @RolesAllowed({"VOLUNTEER", "CONTACT_PERSON"})
+    @PostMapping("/editPassword")
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    public void editProfilePassword(@RequestParam("newPassword") String newPassword,
```
ibaiul: We should encourage the use of the HTTP body to transfer information whenever possible. So it is better to model a DTO instead of using the request parameters.

backend/src/main/java/com/huellapositiva/domain/actions/UpdatePasswordAction.java
```diff
@@ -1,30 +1,34 @@
...
-    EmailCommunicationService emailCommunicationService;
+    protected EmailCommunicationService emailCommunicationService;
 
     @Autowired
-    private JpaCredentialRepository jpaCredentialRepository;
+    protected JpaCredentialRepository jpaCredentialRepository;
 
     @Autowired
-    private final PasswordEncoder passwordEncoder;
+    protected final PasswordEncoder passwordEncoder;
```
ibaiul: private final

backend/src/main/java/com/huellapositiva/domain/actions/UpdatePasswordAction.java
```diff
@@ -62,4 +66,28 @@ public void executePasswordChanging(String hashRecoveryPassword, String password
...
+        if (!passwordEncoder.matches(oldPassword,jpaCredential.getHashedPassword())) {
+            throw new NonMatchingPasswordException("The old password inserted does not match with the one stored in the system");
+        } else if (passwordEncoder.matches(newPassword,jpaCredential.getHashedPassword())) {
+            throw new InvalidNewPasswordException("The new password it exactly the same as the old password");
+        } else {
+            jpaCredentialRepository.updatePassword(newPasswordHash.toString(), email);
+
+            EmailAddress emailAddress = EmailAddress.from(jpaCredential.getEmail());
+            emailCommunicationService.sendConfirmationPasswordChanged(emailAddress);
+        }
```
ibaiul: To improve readability move the else logic out of the if. Structure of code should follow:
```
if (exceptional case) {
   throw exception
}

normal case
```

juanmaMacGyverCode: Ok, it's fixed right now.

backend/src/main/java/com/huellapositiva/domain/actions/UpdatePasswordAction.java
juanmaMacGyverCode: Ok, fixed right now.

backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
juanmaMacGyverCode: Good morning.

I understand your point of view, but the profile is not finished yet, it would be possible to return the user's profile information instead of a No_Content. Which for now is not possible because that endpoint is not done yet. I have it written to refactor that.

ibaiul: The response status is unrelated to the fact that we should use the HTTP body to transmit data.

My comment is not for the response body but for the request body. Avoid using request parameters to receive the newPassword, etc, use a dto instead and send it as a json body to the backend

juanmaMacGyverCode: Ok, i understand, i must to use a dto.

backend/src/main/java/com/huellapositiva/application/controller/HandlerPasswordApiController.java
```diff
@@ -76,4 +87,38 @@ public void sendEmailRecovery(@RequestParam("email") String email){
...
+                            description = "Internal server error, could not fetch the user data due to a connectivity issue."
+                    )
+            }
+    )
+    @RolesAllowed({"VOLUNTEER", "CONTACT_PERSON"})
+    @PostMapping("/editPassword")
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    public void editProfilePassword(@RequestPart("dto") MultipartFile dtoMultipart,
+                                    @AuthenticationPrincipal String email) throws IOException {
+        ChangePasswordDto dto = objectMapper.readValue(dtoMultipart.getBytes(), ChangePasswordDto.class);
```
ibaiul: Multipartfile is meant to be used when you are transferring a file not plain json.
See "registerESALAsReviser" method in "ESALApiController" and do not forget to add the required validation annotation to the request DTO.

juanmaMacGyverCode: About the validation, you mean this:

image

ibaiul: With @NotEmpty should be enough, make sure to create tests for invalid payload as well (HTTP 400). 

The Schema annotations are for Swagger documentation although I believe they should not be required.

juanmaMacGyverCode: In the test invalid payload (HTTP 400), should i check if the data is well-format (only characteres or numbers and minimun 6 for the length, for example) or do you mean if the DTO has its properties with some data? In the second case, with @AllArgsConstructor is solved.

ibaiul: You can perform the validation in the layer you prefer, at controller level or at domain level.

If you want to be consistent with the current approach, perform basic validation like @NotEmpty_ at controller and leave the rest for the domain logic. Consistency aside, validating as much as you can earliest at the controller level using annotations like @Min_, @Size_, regex and others is often used to follow the principle of failing fast and ensuring the data handled by the components is safe in terms of being null-free for instance.

@AllArgsConstructor_ does not perform any validation, it just creates a constructor that takes all the properties as parameters.

juanmaMacGyverCode: Add the validation

backend/src/test/java/com/huellapositiva/api/HandlerPasswordApiControllerShould.java
```diff
@@ -0,0 +1,275 @@
...
+                .header(HttpHeaders.AUTHORIZATION, "Bearer " + jwtResponseDto.getAccessToken())
+                .content(objectMapper.writeValueAsString(new ChangePasswordDto("MíÑewpâssw0rd", DEFAULT_PASSWORD)))
+                .with(csrf())
+                .param("email",DEFAULT_EMAIL)
+                .contentType(APPLICATION_JSON))
+                .andExpect(status().isBadRequest());
+    }
+
+    @Test
+    void return_400_when_the_old_password_does_not_match_the_regular_expression() throws Exception {
```
ibaiul: Create a parameterized test asserting all possible violations: sending null, empty, less than minimum length, ...

See: https://www.baeldung.com/parameterized-tests-junit-5#6-method

backend/src/main/java/com/huellapositiva/application/dto/ChangePasswordDto.java
```diff
@@ -0,0 +1,25 @@
...
+@AllArgsConstructor
+public class ChangePasswordDto {
+
+    @NotEmpty
+    @Size(min = 6, max = 15)
+    @Pattern(regexp = "^[a-zA-Z0-9]*$", message = "Nueva contrasenna falla")
+    private final String newPassword;
+
+    @NotEmpty
+    @Size(min = 6, max = 15)
```
ibaiul: Why wouldn't we allow more than 15 characters?
Why don't we allow special characters? Passwords would be more secure and much much less vulnerable to brute force attacks

juanmaMacGyverCode: Ok, i change it right now. 

The special characters to add "-.,:;+-*/!@#$&" was right? there wasnt problems with accents? (áàäâã)

juanmaMacGyverCode: Done

backend/src/test/java/com/huellapositiva/api/HandlerPasswordApiControllerShould.java
juanmaMacGyverCode: Done

backend/src/main/java/com/huellapositiva/application/dto/ChangePasswordDto.java
```diff
@@ -0,0 +1,28 @@
...
+    @NotEmpty(message = "You must to write something")
+    @NotBlank(message = "The password is not blank space")
+    @Size(min = 6, message = "The password is too short")
+    @Pattern(regexp = "^[a-zA-Z0-9àèìòùÀÈÌÒÙáéíóúýÁÉÍÓÚÝâêîôûÂÊÎÔÛãñõÃÑÕäëïöüÿÄËÏÖÜŸ.,:;+*/|%!@#$&€=<>() -]*$", message = "The new password does not match with the regular expresion")
+    private final String newPassword;
+
+    @NotEmpty(message = "You must to write something")
+    @NotBlank(message = "The password is not blank space")
+    @Size(min = 6, message = "The password is too short")
+    @Pattern(regexp = "^[a-zA-Z0-9àèìòùÀÈÌÒÙáéíóúýÁÉÍÓÚÝâêîôûÂÊÎÔÛãñõÃÑÕäëïöüÿÄËÏÖÜŸ.,:;+*/|%!@#$&€=<>() -]*$", message = "The old password does not match with the regular expresion")
```
ibaiul: I would not check this in the "oldPassword" property. You never know what the previously validation rules were and it might not be backwards compatible.

Image in the future we decide that the minimum length should be 5, and some time later that it should be 4. How would you distinguish that the user trying to update their password had one restriction or the other?

You should only validate the new password upon the current rules.

backend/src/test/java/com/huellapositiva/api/HandlerPasswordApiControllerShould.java
```diff
@@ -0,0 +1,208 @@
...
+
+        // WHEN + THEN
+        mvc.perform(post(baseUri + "/editPassword/")
+                .header(HttpHeaders.AUTHORIZATION, "Bearer " + jwtResponseDto.getAccessToken())
+                .content(objectMapper.writeValueAsString(changePasswordDto))
+                .with(csrf())
+                .param("email",DEFAULT_EMAIL)
+                .contentType(APPLICATION_JSON))
+                .andExpect(status().isBadRequest());
+    }
```
ibaiul: Nice

backend/src/main/java/com/huellapositiva/application/dto/ChangePasswordDto.java
```diff
@@ -0,0 +1,28 @@
...
+
+@Data
+@Builder
+@AllArgsConstructor
+public class ChangePasswordDto {
+
+    @NotEmpty(message = "You must to write something")
+    @NotBlank(message = "The password is not blank space")
+    @Size(min = 6, message = "The password is too short")
+    @Pattern(regexp = "^[a-zA-Z0-9àèìòùÀÈÌÒÙáéíóúýÁÉÍÓÚÝâêîôûÂÊÎÔÛãñõÃÑÕäëïöüÿÄËÏÖÜŸ.,:;+*/|%!@#$&€=<>() -]*$", message = "The new password does not match with the regular expresion")
```
ibaiul: No need for tilde characters. At the time of creating one account you are normally only asked to provide common letters, numbers and common characters.

Consider https://docs.oracle.com/cd/E11223_01/doc.910/e11197/app_special_char.htm#MCMAD416

backend/src/main/java/com/huellapositiva/application/dto/ChangePasswordDto.java
JeremyTrujillo: The old password does not check size and pattern anymore.

backend/src/main/java/com/huellapositiva/application/dto/ChangePasswordDto.java
JeremyTrujillo: We have changed the regex according to oracle.

backend/src/main/java/com/huellapositiva/application/dto/ChangePasswordDto.java
```diff
@@ -0,0 +1,26 @@
...
+import javax.validation.constraints.Pattern;
+import javax.validation.constraints.Size;
+
+@Data
+@Builder
+@AllArgsConstructor
+public class ChangePasswordDto {
+
+    @NotEmpty(message = "You must to write something")
+    @NotBlank(message = "The password is not blank space")
```
ibaiul, line 18: If minimum size is 6 and the space character is not allowed, then it cannot be blank by definition.

ibaiul (approved):

PR #60, opened 2020-12-02
:adhesive_bandage: Changed CSRF configuration

javidiazdom:
### CSRF fix

Based on this [ documentation ](https://owasp.org/www-community/attacks/csrf#:~:text=Cross%2DSite%20Request%20Forgery%20(CSRF,which%20they're%20currently%20authenticated ) and the need of the front-end library used for querying the back end for this XSRF-TOKEN cookie to be accesible from JavaScriopt, I suggest changing this Security configuration.

### DIFF
--- a/backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
+++ b/backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
@@ -73,7 +73,7 @@ class SecurityConfig extends WebSecurityConfigurerAdapter {
         http.cors().and()
                 .csrf()
                 .ignoringAntMatchers("/api/v1/authentication/login", "/api/v1/volunteers", "/api/v1/contactperson")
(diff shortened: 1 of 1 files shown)


juanmaMacGyverCode (approved):

backend/src/main/java/com/huellapositiva/infrastructure/SecurityConfig.java
```diff
@@ -73,7 +73,7 @@ protected void configure(HttpSecurity http) throws Exception {
         http.cors().and()
                 .csrf()
                 .ignoringAntMatchers("/api/v1/authentication/login", "/api/v1/volunteers", "/api/v1/contactperson")
```
ibaiul, line 75: Why do we ignore these paths? A CSRF token should be mandatory in any state changing request unless we have a good reason not to.

PR #61, opened 2020-12-06
Set sonar branch name in sonar pipeline stage

ibaiul:


manumoranchel (approved):

pedroamador (commented):
Cool, I agree with this change.

Maybe we can add sonar check stage also to the front repository

pedroamador (approved):

PR #67, opened 2021-01-26
Fix/jwt subject

ibaiul:
CONTEXT
We cannot use the email of a user as the subject of JWT tokens. The email can be changed and that causes issues with already issued access tokens which do not map an existing user/email any more. That means that currently the only safe way to change the email of a user is to logout that user and force it to login providing the new email.

We will overcome that limitation by using the account ID of the user as the subject, which is precisely the ID of the Credential and does not ever change.

Therefore we should stop retrieving users by email and use the Account ID instead.

I will not extend but there could also be many more misbehaviours including data corruption if we do not have a unique, reliable, non-changing and safe way to handle users.


backend/src/main/java/com/huellapositiva/domain/service/ProfileService.java
```diff
@@ -48,37 +49,37 @@
      * This method update the user profile information in database
      *
      * @param updateProfileRequestDto New user profile information to update
-     * @param email Email of user logged
+     * @param accountId Account ID of user logged
      */
-    public void updateProfile(UpdateProfileRequestDto updateProfileRequestDto, String email) {
-        boolean isNewEmail = !email.equalsIgnoreCase(updateProfileRequestDto.getEmail());
+    public UpdateProfileResult updateProfile(UpdateProfileRequestDto updateProfileRequestDto, String accountId) {
+        JpaVolunteer jpaVolunteer1 = jpaVolunteerRepository.findByAccountIdWithCredentials(accountId).orElse(null);
```
juanmaMacGyverCode: Esta linea no está siendo usada.

backend/src/main/resources/application.properties
```diff
@@ -10,6 +10,8 @@ spring.datasource.password=12345
 spring.datasource.driver-class-name=org.postgresql.Driver
 spring.jpa.show-sql=true
 
+spring.flyway.locations=classpath:db/migration/common,classpath:db/migration/{vendor}
+
```
juanmaMacGyverCode, line 14: ¿Podrías explicarme el por qué? Es que no se que implica. Gracias.

ibaiul: El problema es que H2, la BBDD que usamos en memoria, no soporta las mismas instrucciones que PostgreSQL. 

Para ello tendremos que tener los scripts de migracion comunes en un directorio (db/migration/common) y los especificos en otro de manera que Spring va a elegir automaticamente que directorios usar dependiendo del tipo de BBDD que usamos al arracar la applicatcion.

Si le decimos a spring -> spring.datasource.platform=postgres -> usara:
db/migration/common y db/migration/postgres

Si le decimos a spring -> spring.datasource.platform=h2 -> usara:
db/migration/common y db/migration/h2

juanmaMacGyverCode (dismissed):

PR #74, opened 2021-02-02
Refactored cancelProposal Method

JesusArtiles:
Añadimos el campo cancelReason y además en el action comprobamos que si la proposal está a finished o inadequate no es posible cancelarla.


backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -437,6 +437,10 @@ public ProposalResponseDto fetchProposalWithVolunteers(@PathVariable String idPr
                             responseCode = "204",
                             description = "No Content, proposal status changed to CANCELLED successfully."
                     ),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Bad Request, illegal status."
+                    ),
```
ibaiul: Bar request != illegal status

That the proposal cannot be cancelled does not mean the request is malformed. The request was indeed valid so therefore we should return a more suitable status code like 403, 409, 412

backend/src/main/java/com/huellapositiva/domain/actions/CancelProposalAction.java
```diff
@@ -14,19 +17,25 @@
...
      *
      * (method "updateProposalStatus" returns 0 in case it doesn't find any proposal with that id)
      */
-    public void executeByReviser(String id) {
+    public void executeByReviser(String id, ProposalCancelReasonDto dto) {
+        JpaProposal proposal = jpaProposalRepository.findByNaturalId(id).orElseThrow(EntityNotFoundException::new);
+        Integer status = proposal.getStatus().getId();
+        if(status.equals(ProposalStatus.FINISHED.getId()) || status.equals(ProposalStatus.INADEQUATE.getId()) ){
+            throw new IllegalStateException();
+        }
```
ibaiul: Spacing not correct. Ctrl+Alt+L

JeremyTrujillo: Fixed.

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
JeremyTrujillo: We changed it to 412.

ibaiul (approved):

PR #75, opened 2021-02-04
Revision/submit proposal revision

juanmaMacGyverCode:
Quería comentar una cosa. 

Hay aún un endpoint que todavía no ha sido añadido (editar/revisar la proposal).  Por lo que me surje duda en ProposalApiController.java entre las líneas 332-334. Aún no está definida esa URI por la cual nuestro CONTACT_PERSON puede modificar la proposal, por lo que puede sufrir cambios (aparte que me he dado cuenta que el código esta incorrectamente formateado).

¿La duda es si la URI debería redigirir a la proposal para revisar, o al formulario para editar?

Saludos.

EDIT: Aparte de la revisión que quería darle, que era simplemente cambiar el estado en base de datos y añadir una validación. También cogí y revisé a fondo todo el método, refactoricé los tests y añadí tests Mock para el email. También hice un poco de boy scout con algunos imports que no se usaban.


backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -310,6 +311,10 @@ public ListedProposalsDto fetchListedProposals(@PathVariable Integer page, @Path
                             responseCode = "200",
                             description = "Ok, email with proposal sent to reviser."
                     ),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Bad request. The ID is not in review pending."
+                    ),
```
ibaiul: 400 should only be returned when you send a malformed/invalid payload (HTTP request body).
Use 403, 409, 412. If you want to be consistent with other features return 412 although might not be the most appropriate

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -328,6 +333,8 @@ public void submitProposalRevision(@PathVariable String id,
                     .path(PATH_ID).buildAndExpand(id)
                     .toUri();
             submitProposalRevisionAction.execute(id, dto, uri, accountId);
+        } catch (InvalidStatusIdException ex) {
```
ibaiul: Did you meant InvalidProposalStatusException?

backend/src/main/java/com/huellapositiva/domain/model/valueobjects/ProposalRevisionEmail.java
```diff
@@ -26,7 +24,7 @@
 
     private final Reviser reviser;
 
-    private final Boolean hasFeedback;
```
ibaiul: I never understood why hasFeedback is a field instead of method returning feedback != null

The logic around hasFeedback is extremelly meshy and probably wrong. Why do we need to have that as a field and not a method?

backend/src/main/java/com/huellapositiva/domain/service/ProposalService.java
```diff
@@ -38,4 +58,51 @@ public void enrollVolunteer(String proposalId, Volunteer volunteer) {
...
+
+        if (ProposalStatus.REVIEW_PENDING.getId() != proposal.getStatus().getId()) {
+            throw new InvalidStatusIdException();
+        }
+
+        ESAL esal = proposalRepository.fetch(proposalId).getEsal();
+        ContactPerson contactPerson = contactPersonRepository.findByJoinedEsalId(esal.getId().toString());
+        Reviser reviser = Reviser.from(credentialsRepository.findReviserByAccountId(accountId));
+
+        ProposalRevisionEmail revisionBuilder = ProposalRevisionEmail.builder()
```
ibaiul: revisionBuilder is not accurate, proposalRevisionEmail is.

backend/src/main/java/com/huellapositiva/domain/service/ProposalService.java
```diff
@@ -38,4 +58,51 @@ public void enrollVolunteer(String proposalId, Volunteer volunteer) {
...
+
+    /**
+     * Fetch the ESAL of the database for get the ContactPerson, change status of the proposal and send an email with the revision of the reviser.
+     *
+     * @param proposalId : The id of the proposal to be revised.
+     * @param revisionDto : Contains the email reviser and the feedback if has it.
+     * @param proposalURI : URI the proposal to revise.
+     * @param accountId Account ID of logged user
+     */
+    public ProposalRevisionEmail proposalChangeStatusToChangedRequested(
```
ibaiul: We are in the proposalService, so no need to mention it in every name, it can be deduced from the context. How about?
```
public ProposalRevisionEmail requestChanges(...) {
}
```

backend/src/test/java/com/huellapositiva/util/TestData.java
```diff
@@ -327,6 +327,10 @@ public JpaProposal registerESALAndNotPublishedProposal() {
         return registerESALAndProposal(UNPUBLISHED);
     }
 
+    public JpaProposal registerESALAndReviewPendingProposal() {
+        return registerESALAndProposal(REVIEW_PENDING);
+    }
+
```
ibaiul: registerESALAndReviewXXXXXXProposal() methods are all redundant. Why not use registerESALAndProposal(XXXXXX) directly in the tests instead?

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
juanmaMacGyverCode: Ok, I have written responseCode="412", I think it is the most appropriate.

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
juanmaMacGyverCode: It's done.

backend/src/main/java/com/huellapositiva/domain/model/valueobjects/ProposalRevisionEmail.java
juanmaMacGyverCode: Me neither understood it... If there are message(feedback), then it has implicit hasFeedback. I would remove it. 

In my opinion, which frontend does the string.length... hasFeedback does not contribute anything. Feedback should be optional (String).

Do I remove hasFeedback? Is extremelly meshy...

backend/src/main/java/com/huellapositiva/domain/service/ProposalService.java
juanmaMacGyverCode: It's done

backend/src/main/java/com/huellapositiva/domain/service/ProposalService.java
juanmaMacGyverCode: It's done.

backend/src/test/java/com/huellapositiva/util/TestData.java
juanmaMacGyverCode: I dont know, I saw that it was like this. It's refactored right now.

backend/src/main/java/com/huellapositiva/domain/model/valueobjects/ProposalRevisionEmail.java
```diff
@@ -2,16 +2,14 @@
...
-import lombok.Builder;
-import lombok.Data;
-import lombok.RequiredArgsConstructor;
-import lombok.SneakyThrows;
+import lombok.*;
 
 import java.net.URI;
 
 @RequiredArgsConstructor
 @Data
```
JeremyTrujillo: Data already includes @RequiredArgsConstructor, so it's redundant to use the other annotation.

backend/src/test/java/com/huellapositiva/api/ProposalControllerShould.java
```diff
@@ -550,12 +543,16 @@ void return_200_when_submitting_revision_as_revisor_has_feedback_but_not_feedbac
                 .contentType(MediaType.APPLICATION_JSON)
                 .accept(MediaType.APPLICATION_JSON))
                 .andExpect(status().isOk());
+
+        JpaProposal jpaProposalWithStatusChanged = jpaProposalRepository.findByNaturalId(proposalId)
+                .orElseThrow(() -> new UserNotFoundException("Proposal not found. Account ID: " + proposalId));
+        assertThat(jpaProposalWithStatusChanged.getStatus().getId()).isEqualTo(2);
```
JeremyTrujillo: Instead of that magic number, we should get the id from the ProposalStatus enum. That way, if we need to change the order in the future, we will still compare this to CHANGES_REQUESTED, and its easier to understand.

backend/src/main/java/com/huellapositiva/domain/actions/ResendEmailConfirmationAction.java
```diff
@@ -38,7 +38,7 @@ public void execute(String accountId) {
         JpaCredential jpaCredential = jpaCredentialRepository.findByAccountId(accountId)
                 .orElseThrow(() -> new UsernameNotFoundException("User with account ID: " + accountId + " was not found."));
 
-        if (jpaCredential.getEmailConfirmed()) {
+        if (Boolean.TRUE.equals(jpaCredential.getEmailConfirmed())) {
```
JeremyTrujillo, line 41: I suppose you made this to avoid using a Boolean object in the if. Would it not be better to use the .booleanValue() method of Boolean to get the primitive instead of an equals()?

backend/src/main/java/com/huellapositiva/domain/service/ProposalService.java
```diff
@@ -38,4 +58,51 @@ public void enrollVolunteer(String proposalId, Volunteer volunteer) {
...
+            URI proposalURI,
+            String accountId) {
+
+        Proposal proposal = proposalRepository.fetch(proposalId);
+
+        if (ProposalStatus.REVIEW_PENDING.getId() != proposal.getStatus().getId()) {
+            throw new InvalidProposalStatusException();
+        }
+
+        ESAL esal = proposalRepository.fetch(proposalId).getEsal();
```
JeremyTrujillo: You have previously fetched the proposal, there's no need to do it again.

backend/src/test/java/com/huellapositiva/unit/ProposalRevisionShould.java
```diff
@@ -0,0 +1,79 @@
...
+                .reviser(reviser)
+                .token(Token.createToken())
+                .build();
+
+        when(proposalService.requestChanges(proposalID, proposalRevisionDto, uri, accountID))
+                .thenReturn(proposalRevisionEmail);
+
+        submitProposalRevisionAction.execute(proposalID, proposalRevisionDto, uri, accountID);
+
+        verify(emailCommunicationService).sendSubmittedProposalRevision(any());
```
JeremyTrujillo: You have created a ProposalRevisionEmail to be returned by your proposalService mock, and that email should me used by EmailCommunicationService. Then, why do you verify it using any() instead of your own email?

JeremyTrujillo (changes requested):

backend/src/main/java/com/huellapositiva/domain/model/valueobjects/ProposalRevisionEmail.java
juanmaMacGyverCode: It's done

backend/src/test/java/com/huellapositiva/api/ProposalControllerShould.java
juanmaMacGyverCode: It's done.

backend/src/main/java/com/huellapositiva/domain/actions/ResendEmailConfirmationAction.java
juanmaMacGyverCode: SonarLint rules!

backend/src/main/java/com/huellapositiva/domain/model/valueobjects/ProposalRevisionEmail.java
ibaiul: I would. Just make it a method that returns feedback!=null

juanmaMacGyverCode: It's done.

backend/src/main/java/com/huellapositiva/domain/service/ProposalService.java
juanmaMacGyverCode: Ok, it's done.

backend/src/test/java/com/huellapositiva/unit/ProposalRevisionShould.java
juanmaMacGyverCode: Fixed it

JeremyTrujillo (approved):

PR #82, opened 2021-02-24
Feature/change published proposal to enrollment closed

ManuPerezDev:
This is a feature so that the contact person can update the proposal status from PUBLISHED to ENROLLMENT_CLOSED.

PR #83, opened 2021-02-25
Feature/change enrollment closed proposal to published

ManuPerezDev:
The contact person now can change the enrollment_closed proposal status to published.
We have doubts about the endpoint naming because we already have /publish endpoint. We think that we could add the expected status thus: /publish/enrollment-closed


ibaiul:
Merged into PR #82

PR #86, opened 2021-03-03
Feature/update proposal image

JesusArtiles:
Endpoint para actualizar la foto de una proposal y enviar email para notificarlo.


backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -544,5 +553,52 @@ public void changeStatusVolunteerInProposal(@RequestBody List<ChangeStatusVolunt
...
         changeReviewPendingProposalToPublishedAction.execute(dto.getIdProposal());
     }
-}
 
+    @Operation(
+            summary = "Updates the proposal image",
+            description = "The contact person can update the image of the proposal.",
+            tags = {"proposals, contact person"},
+            parameters = {
+                    @Parameter(name = "X-XSRF-TOKEN", in = ParameterIn.HEADER, required = true, example = "ff79038b-3fec-41f0-bab8-6e0d11db986e", description = "For taking this value, open your inspector code on your browser, and take the value of the cookie with the name 'XSRF-TOKEN'. Example: a6f5086d-af6b-464f-988b-7a604e46062b"),
```
ibaiul: CSRF token is sent as a cookie and a query parameter. Check the Postman collection in the project from the usage

`ParameterIn.QUERY`

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -544,5 +553,52 @@ public void changeStatusVolunteerInProposal(@RequestBody List<ChangeStatusVolunt
...
+                            responseCode = "403",
+                            description = "Forbidden, The contact person related to this proposal does not match the logged contact person.."
+                    ),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Internal server error, could not fetch the user data due to a connectivity issue."
+                    )
+            }
+    )
+    @PostMapping("/udpateProposalImage")
```
ibaiul: PUT /{id}/image -> RESTful URLs

where {id} would be the proposalId

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
```diff
@@ -544,5 +553,52 @@ public void changeStatusVolunteerInProposal(@RequestBody List<ChangeStatusVolunt
...
+    @RolesAllowed("CONTACT_PERSON")
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    public void updateProposalImage(@RequestPart("photo") MultipartFile photo, @RequestBody String proposalId,
```
ibaiul: porposalId is provided in the URI

backend/src/main/java/com/huellapositiva/domain/actions/UpdateProposalImageAction.java
```diff
@@ -0,0 +1,73 @@
...
+    @Autowired
+    private JpaContactPersonRepository contactPersonRepository;
+
+    @Autowired
+    private ImageService imageService;
+
+    private final RemoteStorageService remoteStorageService;
+
+    public void execute(MultipartFile photo, String accountId, String proposalId) throws IOException {
+        imageService.validateProfileImage(photo);
```
ibaiul: Validate profile image?

backend/src/main/java/com/huellapositiva/domain/actions/UpdateProposalImageAction.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private void validateProposalUpdate(JpaContactPerson accountContactPerson, JpaProposal jpaProposal) throws AccessDeniedException {
+        JpaContactPerson proposalContactPerson = contactPersonRepository.findByEsalId(jpaProposal.getEsal().getId()).orElseThrow(EntityNotFoundException::new);
+        if(!accountContactPerson.getCredential().getEmail().equals(proposalContactPerson.getCredential().getEmail())){
+            throw new AccessDeniedException("The contact person related to this proposal does not match the logged contact person.");
+        }
+        if(!jpaProposal.getStatus().getId().equals(ProposalStatus.PUBLISHED.getId()) &&
+                !jpaProposal.getStatus().getId().equals(ProposalStatus.REVIEW_PENDING.getId())){
+            throw new IllegalStateException();
```
ibaiul: The exception is lacking a proper message explaining the reason of the failure

backend/src/main/java/com/huellapositiva/domain/actions/UpdateProposalImageAction.java
```diff
@@ -0,0 +1,73 @@
...
+        proposal.setImage(url);
+        proposal.setStatus(ProposalStatus.REVIEW_PENDING);
+        proposalRepository.save(proposal);
+
+        emailCommunicationService.sendProposalImageUpdateEmail(EmailAddress.from(accountContactPerson.getCredential().getEmail()));
+    }
+
+    private void validateProposalUpdate(JpaContactPerson accountContactPerson, JpaProposal jpaProposal) throws AccessDeniedException {
+        JpaContactPerson proposalContactPerson = contactPersonRepository.findByEsalId(jpaProposal.getEsal().getId()).orElseThrow(EntityNotFoundException::new);
+        if(!accountContactPerson.getCredential().getEmail().equals(proposalContactPerson.getCredential().getEmail())){
```
ibaiul: We cannot compare users using a mutable property such as the email address, which is subject to be changed when updating the profile.

The users are not really identified by their email address, they are identified by ..., surprise, surprise, ..., their IDs. You simply need to check that the `accountId` of the logged user matches the one you are interested into.


ContactPerson ID3, user@example.com -> ContactPerson ID3, user2@example.com
User with ID=3 can update its email address, it is just a detail of its profile that is used for login (we could use other fields too) and communication purposes

ContactPerson user@example.com, ID3  -> ContactPerson user@example.com, ID4 (conflict)
User with email address user@example.com cannot update its ID, the ID is immutable as it is the unique identifier of that user. That would lead to the creation of two users with the same email address which the database would not allow

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
JesusArtiles: Solved

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
JesusArtiles: Solved

backend/src/main/java/com/huellapositiva/domain/actions/UpdateProposalImageAction.java
JesusArtiles: Ok, we created a new method to validate proposal images

backend/src/main/java/com/huellapositiva/domain/actions/UpdateProposalImageAction.java
JesusArtiles: Now it has one

backend/src/main/java/com/huellapositiva/domain/actions/UpdateProposalImageAction.java
JesusArtiles: You are right, now we use getCredential.getId to compare

backend/src/main/java/com/huellapositiva/application/controller/ProposalApiController.java
JeremyTrujillo: Done

ibaiul (approved):

PR #87, opened 2021-03-04
Feature/edit esal

JeremyTrujillo:
Esta rama consiste en el endpoint para modificar los datos de una esal.


backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
```diff
@@ -99,6 +104,52 @@ public void registerESAL(@Validated @RequestBody ESALRequestDto dto,
         }
     }
 
+    @Operation(
+            summary = "Updates the information of an ESAL",
+            description = "Updates the information of the ESAL linked to the logged employee. Roles allowed CONTACT_PERSON and CONTACT_PERSON_NOT_CONFIRMED.",
+            tags = "ESAL",
+            parameters = {
+                    @Parameter(name = "X-XSRF-TOKEN", in = ParameterIn.HEADER, required = true, example = "a6f5086d-af6b-464f-988b-7a604e46062b", description = "For take this value, open your inspector code on your browser, and take the value of the cookie with the name 'XSRF-TOKEN'. Example: a6f5086d-af6b-464f-988b-7a604e46062b"),
```
ibaiul: CSRF token is not passed as a header, but as a query parameter. See Postman collection in the project for usage

backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
```diff
@@ -99,6 +104,52 @@ public void registerESAL(@Validated @RequestBody ESALRequestDto dto,
...
+                            description = "Conflict, the provided name is already taken."
+                    ),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Internal server error, could not update the esal.",
+                            content = @Content(mediaType = "application/json")
+                    )
+            }
+    )
+    @PostMapping("/update")
```
ibaiul: @_PutMapping("/") -> as a general rule we use POST for creation and PUT for updates

backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
```diff
@@ -99,6 +104,52 @@ public void registerESAL(@Validated @RequestBody ESALRequestDto dto,
...
+    @RolesAllowed({"CONTACT_PERSON", "CONTACT_PERSON_NOT_CONFIRMED"})
+    public void updateESAL(@Validated @RequestBody UpdateESALDto dto,
+                           @Parameter(hidden = true) @AuthenticationPrincipal String accountId){
+        try{
+            updateESALAction.execute(dto, accountId);
+        } catch (DataIntegrityViolationException ex){
+            throw new ResponseStatusException(HttpStatus.CONFLICT, "ESAL named " + dto.getName() + " already exists.");
+        } catch (UserNotFoundException ex){
+            throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, "Could not find the user due to a connectivity issue.");
```
ibaiul: UserNotFound and internal server error are two completely different things.

Not found is not a server error, it is a client error indeed which occurred due to a client having requested something that did not exist

Think which of the two fits your use case

backend/src/main/java/com/huellapositiva/domain/actions/UpdateESALAction.java
```diff
@@ -0,0 +1,52 @@
...
+        JpaLocation location = JpaLocation.builder()
+                .surrogateKey(esalLocation.getSurrogateKey())
+                .id(esalLocation.getId())
+                .island(dto.getIsland())
+                .zipCode(dto.getZipCode())
+                .province(dto.getProvince() != null ? dto.getProvince() : null)
+                .town(dto.getTown() != null ? dto.getTown() : null)
+                .address(dto.getAddress() != null ? dto.getAddress() : null)
+                .build();
+        esal.setLocation(location);
```
ibaiul: `esalLocation` is an object loaded from the database, you should update it and persist it.

`location` object should not be created

```
esalLocation.setIsland(dto.getIsland());
esalLocation.set...
...
jpaLocationRepository.save(esalLocation);
```

backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
JesusArtiles: Solved

backend/src/main/java/com/huellapositiva/domain/actions/UpdateESALAction.java
JesusArtiles: We did not use the save for location only, we prefer using jpaESALRepository save for updating both at the same time.

backend/src/main/java/com/huellapositiva/domain/actions/UpdateESALAction.java
```diff
@@ -0,0 +1,48 @@
...
+        esal.setWebsite(dto.getWebsite());
+        esal.setRegisteredEntity(dto.isRegisteredEntity());
+        esal.setEntityType(dto.getEntityType());
+    }
+
+    private void updateLocation(JpaESAL esal, UpdateESALDto dto) {
+        JpaLocation esalLocation = esal.getLocation();
+        esalLocation.setIsland(dto.getIsland());
+        esalLocation.setZipCode(dto.getZipCode());
+        esalLocation.setProvince(dto.getProvince() != null ? dto.getProvince() : null);
```
ibaiul:
```
esaLocation.setProvince(dto.getProvince());
```

If `dto.getProvince()` is null it will lead to the same result as assigning it to `null` directly

backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
JeremyTrujillo: Currently, we have changed it to throw a HttpStatus.NOT_FOUND for consistency, but we are still unsure about this. This exception is thrown when the system fetches the current user's data. It's not that the user has requested a non-existent user, but that the system cannot get that user's data, although they are logged in. That's why we thought it should throw a server error instead of a client one.

backend/src/main/java/com/huellapositiva/domain/actions/UpdateESALAction.java
JeremyTrujillo: You are right, I don't know why I didn't see that. Changed for province, town and address.

backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
JeremyTrujillo: Done

ibaiul (approved):

PR #88, opened 2021-03-05
Feature/update contact person profile photo

ManuPerezDev:
Esta rama está hecha para proporcionar al Contact Person un endpoint para poder actualizar su imagen de perfil.


backend/src/main/java/com/huellapositiva/application/controller/ESALContactPersonApiController.java
```diff
@@ -83,4 +97,45 @@ public JwtResponseDto registerContactPerson(@Validated @RequestBody RegisterESAL
     public ProposalResponseDto getMember(@PathVariable Integer id) {
         throw new UnsupportedOperationException();
     }
+
+    @Operation(
+            summary = "Upload contact person Photo",
+            description = "Upload contact person Photo to profile",
+            tags = "user",
+            parameters = {
+                    @Parameter(name = "X-XSRF-TOKEN", in = ParameterIn.HEADER, required = true, example = "a6f5086d-af6b-464f-988b-7a604e46062b", description = "For take this value, open your inspector code on your browser, and take the value of the cookie with the name 'XSRF-TOKEN'. Example: a6f5086d-af6b-464f-988b-7a604e46062b"),
```
ibaiul: The CSRF token has to be provided both as a cookie and a query parameter. Originally it was provided through a header but in was changed to satisfy frontend requirements

See Postman collection in the project for its usage
`
ParameterIn.QUERY
`

backend/src/main/java/com/huellapositiva/domain/repository/ContactPersonRepository.java
```diff
@@ -5,20 +5,65 @@
...
+    /**
+     * This method store in DB the URL of user profile photo
+     *
+     * @param contactPerson The contact person information
+     */
+    public void updatePhoto(ContactPerson contactPerson) {
+        JpaContactPerson jpaContactPerson = jpaContactPersonRepository.findByAccountId(contactPerson.getAccountId().toString())
+                .orElseThrow(() -> new NoSuchElementException("No exists contact person with: " + contactPerson.getId()));
+
+        boolean profileIsNull = jpaContactPerson.getContactPersonProfile() == null;
```
ibaiul: Do we really need a variable for that? You can perform the check directly

backend/src/main/java/com/huellapositiva/infrastructure/orm/repository/JpaContactPersonRepository.java
```diff
@@ -35,4 +35,12 @@
...
     @Query("FROM JpaContactPerson cp LEFT JOIN FETCH cp.joinedEsal je WHERE je.id = :id")
     Optional<JpaContactPerson> findByEsalId(@Param("id") String id);
+
+    @Query("FROM JpaContactPerson cp LEFT JOIN FETCH cp.contactPersonProfile p LEFT JOIN FETCH cp.credential c WHERE c.id = :accountId")
+    Optional<JpaContactPerson> findByAccountIdWithProfile(@Param("accountId") String accountId);
+
+    @Modifying
+    @Transactional
+    @Query("UPDATE JpaContactPerson cp SET cp.contactPersonProfile = :profile WHERE cp.id = :id")
+    Integer updateProfile(@Param("id") String id, @Param("profile") JpaContactPersonProfile profile);
```
ibaiul: This query should be removed.

In ORM world the relations are not persisted on that way. You need to have the two objects belonging to the relation loaded in the Hibernate session (first level cache).

That means that if you save the object that owns (contains foreign key in its DB table) the relation Hibernate will handle the rest for you.

```
JpaContactPerson jpaContactPerson = ...
JpaContactPersonProfile jpaContactPersonProfile = '... -> this object must be in the first level cache (loaded from DB or persisted reference)
jpaContactPerson.setProfile(jpaContactPersonProfile);
jpaContactPersonRepository.save(jpaContactPerson); -> Hibernate will see the relation and get the profile id to persist it in the JpaContactPerson table
```

A big plus, it is hard to justify writing production code that is only used in tests. We should not create any production code for testing purposes as a general rule.

backend/src/main/resources/db/migration/common/V1_0_18__Add_Contact_Person_Profile_Photo.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE contact_person_profile
+    ADD COLUMN photo_url varchar (1030);
```
ibaiul, line 2: Where does this number come from? It is just a curiosity

backend/src/test/java/com/huellapositiva/util/TestData.java
```diff
@@ -542,6 +547,24 @@ private JpaVolunteer createVolunteerProfile(String email) {
...
+                .surname("apellidos")
+                .phoneNumber("12412412125")
+                .photoUrl("Una direccion ahi")
+                .build();
+        jpaContactPersonProfileRepository.save(jpaContactPersonProfile);
+
+        JpaContactPerson jpaContactPerson = jpaContactPersonRepository.findByEmail(email).orElseThrow(UserNotFoundException::new);
+        jpaContactPersonRepository.updateProfile(jpaContactPerson.getId(), jpaContactPersonProfile);
+
+        jpaContactPerson.setContactPersonProfile(jpaContactPersonProfile);
```
ibaiul: JPA relations are not persisted in this way. See extended comment above.

```
jpaContactPerson.setProfile(jpaProfile);
jpaContactPerson = jpaContactPersonRepository.save(jpaContactPerson) -> when you save an object using a jpa repository you should reference the object returned by the repository cause it is the only object containing the latest reference from the database (will contain all fields created by the database engine or Hibernate such IDs, creation-timestamp, update-timestamp, ... ). 
```

backend/src/main/java/com/huellapositiva/application/controller/ESALContactPersonApiController.java
ManuPerezDev: You mean change name = "X-XSRF-TOKEN" -> name = "_csrf" and  ParameterIn.Header -> ParameterIn.Query?

backend/src/main/resources/db/migration/common/V1_0_18__Add_Contact_Person_Profile_Photo.sql
ManuPerezDev: Don't remember. I'll ask.

ManuPerezDev: I think It was put it that way because we don't know the url length. But maybe it's better only varchar removing the 1030?

backend/src/test/java/com/huellapositiva/api/ESALContactPersonControllerShould.java
```diff
@@ -187,4 +194,114 @@ void return_400_when_register_contact_person_with_malformed_data_request(Registe
...
+        mvc.perform(multipart("/api/v1/contactperson/profile/photo")
+                .file(new MockMultipartFile("photo", "photo-test.PNG", "image/png", is))
+                .header(HttpHeaders.AUTHORIZATION, "Bearer " + jwtResponseDto.getAccessToken())
+                .contentType(MULTIPART_FORM_DATA)
+                .with(csrf())
+                .accept(APPLICATION_JSON))
+                .andExpect(status().isNoContent());
+
+        contactPerson = jpaContactPersonRepository.findByAccountIdWithProfile(contactPerson.getCredential().getId()).orElseThrow(UserNotFoundException::new);
+        assertThat(contactPerson.getContactPersonProfile().getPhotoUrl()).isNotNull();
```
JeremyTrujillo: In this test, the contact person already has a photoURL before uploading the new one. So, this assert is not enough to ensure that the new image has been uploaded.

ibaiul (approved):

PR #92, opened 2021-03-24
Fetch esal

JeremyTrujillo:
This branches fetches an ESAL's information.

### DIFF
--- a/backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
+++ b/backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
@@ -1,20 +1,19 @@
 package com.huellapositiva.application.controller;
 
 import com.huellapositiva.application.dto.ESALRequestDto;
+import com.huellapositiva.application.dto.GetESALResponseDto;
 import com.huellapositiva.application.dto.UpdateESALDto;
 import com.huellapositiva.application.exception.ESALAlreadyExistsException;
 import com.huellapositiva.application.exception.InvalidFieldException;
 import com.huellapositiva.application.exception.UserNotFoundException;
-import com.huellapositiva.domain.actions.DeleteESALAction;
-import com.huellapositiva.domain.actions.RegisterESALAction;
-import com.huellapositiva.domain.actions.UpdateESALAction;
-import com.huellapositiva.domain.actions.UploadLogoAction;
+import com.huellapositiva.domain.actions.*;
 import com.huellapositiva.domain.exception.EmptyFileException;
 import com.huellapositiva.domain.exception.UserAlreadyHasESALException;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.enums.ParameterIn;
 import io.swagger.v3.oas.annotations.media.Content;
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/huellapositiva/domain/actions/FetchESALAction.java
```diff
@@ -0,0 +1,40 @@
...
+    @Autowired
+    private JpaContactPersonRepository jpaContactPersonRepository;
+
+    public GetESALResponseDto execute(String accountId) {
+        JpaContactPerson jpaContactPerson = jpaContactPersonRepository.findByAccountIdWithProfile(accountId)
+                .orElseThrow(() -> new UserNotFoundException("Contact Person not found. Account ID: " + accountId));
+        JpaESAL joinedEsal = jpaContactPerson.getJoinedEsal();
+
+        GetESALResponseDto.GetESALResponseDtoBuilder builder = GetESALResponseDto.builder();
+        builder.name(joinedEsal.getName())
```
ibaiul: I think this might lead to a NullPointerException in some cases.

I see two flows for creating an ESAL:

1. Flow (no issue)
- Create contact person
- Contact person creates an ESAL and both are linked

2. Flow (multiple issues)
- Reviser creates an ESAL which leads to having an ESAL witthout a ContactPerson
- If you retrieve the ESAL you get a NPE, also there is no apparent way to link a ContactPerson to an ESAL that was created by a reviser.

backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
```diff
@@ -104,6 +105,51 @@ public void registerESAL(@Validated @RequestBody ESALRequestDto dto,
...
+                            content = @Content(mediaType = "application/json")
+                    ),
+                    @ApiResponse(
+                            responseCode = "500",
+                            description = "Internal server error, could not fetch the esal data due to a connectivity issue.",
+                            content = @Content(mediaType = "application/json")
+                    )
+            }
+    )
+    @GetMapping
```
ibaiul: GET /esal/{id} and check that the ContactPerson is the real owner of the ESAL

GET / is reserved to retrieve all resources or a paginated list

Consider that in the future a volunteer should also be able to retrieve public/limited set of information about an ESAL using the same endpoint.

fetchEsalAction.executeAsOwner
fetchEsalAction.executeAsVolunteer

backend/src/main/java/com/huellapositiva/domain/actions/FetchESALAction.java
JeremyTrujillo: Solved, now it receives the ID of an ESAL and we check if the logged contact person is its owner, it doesn't throw NPE anymore.

backend/src/main/java/com/huellapositiva/application/controller/ESALApiController.java
JeremyTrujillo: We have created executeAsOwner for the Contact Person to be able to fetch and edit their ESAL information, but we haven't created other type of fetch yet, as we don't plan to have it included in the MVP.

ibaiul: That makes total sense.

ibaiul (approved):

PR #107, opened 2021-05-19
Add AWS SSM dependency to fix issue with AwsSsmConfig class

manumoranchel:
There was a dependency issue on startup with the SSM configuration. I added such dependency to fix it in my local. Although it can be cause by other reason

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -158,6 +158,11 @@
             <version>${aws.sdk.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


ibaiul (approved):

manumoranchel:
Holding the merge. The issue this PR tries to fix might be do to some local caching issues. More test required before incorporating the dependency

manumoranchel:
After several cleanups and attempts to reproduce the issue this PR is meant to fix, I am unable to reproduce it.
I am closing the PR now

